Issue #7, opened 2026-03-16
Add bf16 support to tile DSL


elibol:
Add bf16 support in the tile DSL/kernel element type system (beyond just host tensor dtype support).

PR #16, opened 2026-03-17
feat: Add bf16 support to tile DSL/kernel element type system

goog00:
Hi, this PR adds bf16 support to the tile DSL/kernel element type system (not just host tensor dtype support) #7 

### What Changed
- Added bf16 as a valid DSL ElementType in _core.rs.
- Extended compiler float handling to recognize bf16.
- Added bf16 support in conversion lowering paths (itof, ftoi, ftof) and constant encoding.
- Enabled bf16 in binary float op lowering.

### Tests
Added/updated compile-time MLIR tests to cover bf16:
- unary_math_ops.rs
- binary_math_ops.rs
- type_conversion_ops.rs

All targeted tests pass.

### DIFF
--- a/cutile/tests/type_conversion_ops.rs
+++ b/cutile/tests/type_conversion_ops.rs
@@ -3,8 +3,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 use cutile;
+use cutile::{api::*, tensor::*, tile_kernel::*};
 use cutile_compiler::compiler::{CUDATileFunctionCompiler, CUDATileModules};
 use cutile_compiler::cuda_tile_runtime_utils::get_gpu_name;
+use half::bf16;
+use std::sync::Arc;
 
 mod common;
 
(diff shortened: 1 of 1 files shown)


cutile/tests/type_conversion_ops.rs
```diff
@@ -164,3 +174,37 @@ fn compile_exti_unsigned() -> () {
...
+            None,
+            gpu_name,
+        )
+        .expect("Failed.");
+        let module_op_str = compiler
+            .compile()
+            .expect("Failed.")
+            .as_operation()
+            .to_string();
+        println!("\n=== BF16 CONVERSION MLIR ===\n{}", module_op_str);
```
elibol, line 226: Can you add an additional test to execute the kernel and make sure the conversion to/from f32 succeeded?

goog00: Thanks for the review! I’ll rebase onto main and add a runtime test for the bf16 <-> f32 conversion

goog00: Thanks for the suggestion! I’ve added runtime tests for bf16 <-> f32 conversion, including round-trip, bf16 -> f32, and f32 -> bf16 cases

PR #19, opened 2026-03-17
Stricter clippy

ur4t:


elibol:
@ur4t I think this is probably good to go, but I'm going to review this once we have clippy up and running as part of the CI. Thanks for the contribution!

ur4t:
@elibol Now this PR is ready to go. All tests, examples and benchmarks work as expected.

> Besides, I have find some benchmarks such as FMHA are broken on my machine even in main branch.

`load_from_view_latency` emits `optimization_hints` with `latency` specified, which is not officially supported on all architectures yet. Using `load_from_view` instead or patched `cuda-tile`, both work well.

elibol:
@ur4t thanks for these. Rather than ask you to rebase across three months of churn, I am carrying the lint policy and the cleanups forward on top of current main in a follow-up PR, with you credited as co-author, and closing these two in its favor. I will link the PR here once it is up.

elibol:
Carried forward in #263 (you are credited as co-author on the squash).

Issue #12, opened 2026-03-16
Custom tensor allocator support


elibol:
Support a specific allocator for tensor creation to allow controlling pool allocation.

ptxpapi:
would love to be assigned this issue, if possible.

goog00:
Thank you @elibol and @ptxpapi — this is really helpful, and I think the point about step 2 makes sense.

I agree that if the MVP stays within CUDA’s async memory pool path (e.g. allocation from a specific pool, but still deallocation through `cuMemFreeAsync` / the existing `free_async(...)` path), then changing `DeviceBox::drop()` may not actually be necessary for the first version.

Given that, maybe a narrower mvp for the first version could be:

- confirm the pool-related CUDA APIs/types are available in the current bindings (or extend them if needed),
- add an explicit pool-aware allocation path in `cuda-core`,
- add opt-in tensor creation APIs like `uninitialized_in(...)` / `zeros_in(...)`,
- and keep the current `DeviceBox` drop path unchanged as long as async free remains compatible.

In that case, the provenance/deallocation-mode idea may be more relevant for a broader allocator abstraction later, especially if we want to support non-async allocation paths or explicit pool lifetime management.

The pool ownership / cross-thread drop / user-managed-vs-context-managed pool questions also seem like key scope points here, and probably need a bit more clarification.

I’m happy to help on whichever part makes the most sense once we get that clarification, and also use this as a chance to learn more about this part of the stack.

goog00:
@elibol I'm picking this up now. 
Before I start coding,  i  want to align on one design question that came out of reviewing #26 against the current main (specifically the `DeviceOp` restructuring in #79).

## Context
PR #26's `device_alloc_async` took `(stream, device_id)` as separate parameters, then used `device_id` to look up the pool from the thread-local `AsyncDeviceContext`:
```
pub unsafe fn device_alloc_async(
    num_bytes: usize,
    stream: &Arc,
    device_id: usize,          // ← used to look up pool
) -> Result { ... }
```
This works, but it re-introduces a dependency on `device_id` as a separate parameter- when the whole point of `cudaMallocAsync` design is that the stream already carries device information.  It also means callers must keep `device_id` and `stream` in sync manually.

## Observation
Every `ExecutionContext` is created inside `with_default_device_policy`, which already has access to the `AsyncDeviceContext`:

```rust
with_default_device_policy(|policy| {
    let stream = policy.next_stream()?;
[...]

Issue #13, opened 2026-03-16
Add support for global constants and module-level global memory allocation


elibol:
Add full Rust API support for module-level global memory allocation and global constants using the cuda_tile.global operation. This enables important parallel algorithms which rely on global atomics for signaling.

ptxpapi:
would love to be assigned this issue, if possible.

elibol:
Landed: module-level globals via `Global<E, D>` and `cuda_tile.get_global` came in with #135, `Global::ptr()` was added in #240, and `cutile/tests/gpu/global_counter.rs` exercises the global-atomics signaling case this issue was about. Closing; documenting `Global` in the book is a docs follow-up.

PR #26, opened 2026-03-22
feat: Support context-level memory pool routing for tensor allocation

ghost:
**Closes https://github.com/NVlabs/cutile-rs/issues/12**: *Support a specific allocator for tensor creation to allow controlling pool allocation.*
 
---
 
All tensor allocations currently route through `cuMemAllocAsync` on the operation stream, hitting the device's default async memory pool implicitly. This PR adds an opt-in `CUmemoryPool` configuration at the `AsyncDeviceContext` level, so that all downstream allocation paths (`Tensor::uninitialized`, `copy`, `copy_to_device`, `copy_host_vec_to_device`) transparently route through `cuMemAllocFromPoolAsync` when a pool is configured. No changes to `DeviceBox::drop`, no new public tensor API surface, no generics infection.
 
```rust
// End-user experience: configure once, allocate transparently.
let pool = unsafe { CudaMemPool::new(&cuda_ctx)? };
unsafe { set_device_pool(device_id, Some(pool.handle()))? };
 
// Every tensor allocation now routes through the pool.
let x = api::zeros::<f32>([1024]).await;
let y = api::ones::<f32>([512, 512]).await;
 
// Revert to default cuMemAllocAsync path.
unsafe { set_device_pool(device_id, None)? };
```
 
---
 
### Why context-level, not per-tensor `_in` variants
 
In the [issue discussion](https://github.com/NVlabs/cutile-rs/issues/12#issuecomment-4102975050), the proposed approach was to add `uninitialized_in(len, pool)` / `zeros_in(shape, pool)` variants, mirroring Rust's nightly `Box::new_in` pattern. The following are three evaluated design paths before settling on context-level routing:
 
| Approach | Tradeoff |
|---|---|
| **Per-tensor `_in` variants** | Doubles the public API surface. Pushes pool selection into every call site, which is noisy when most of the allocations target the same pool. |
| **`DeviceAllocator` trait on `DeviceBox<T, A>`** | Leans idiomatic by design, but infects generics through `DeviceBox` → `Tensor` → `Partition` → every function signature. Seems too far for an MVP of this PR, especially without clarity around some questions asked within the issue discussion. |
| **Context-level pool field** *(this PR)* | Zero API surface change to `Tensor` or creation functions. The allocation decision is made once at context init, and `device_operation::with_context` already threads runtime state (stream, device ID) into every operation. |
 
The assumed dominant use case ("training allocations use pool A, inference uses pool B") maps cleanly to context-level configuration. Per-tensor override can be layered on top as a follow-up if needed.
 
### Why `DeviceBox::drop` is unchanged
 
The also-proposed step of extending `DeviceBox` with "provenance" or a "deallocation mode" is unnecessary. `cuMemFreeAsync` handles pool-allocated memory correctly at the driver level as CUDA tracks pool provenance internally and returns memory to the originating pool regardless of which stream calls `cuMemFreeAsync`. The existing `Drop` impl works as-is:
 
```rust
// DeviceBox::drop - unchanged, works for both default and pool-allocated memory.
impl<T: Send + ?Sized> Drop for DeviceBox<T> {
    fn drop(&mut self) {
        unsafe {
            with_deallocator_stream(self._device_id, |stream| {
                free_async(self.cudptr, stream);  // cuMemFreeAsync handles pool provenance
            })
        }
    }
}
```
 
This is the key simplification: by not touching `DeviceBox`, the change surface stays small and the risk of introducing deallocation bugs is zero.
 
---
 
### `cuda-core` — Pool FFI wrappers and RAII type
 
**`memory` module** (`cudarc_shim.rs`): Added `malloc_from_pool_async`, a thin wrapper around `cuMemAllocFromPoolAsync`, alongside the existing `malloc_async`:
 
```rust
pub unsafe fn malloc_from_pool_async(
    stream: sys::CUstream,
    num_bytes: usize,
    pool: sys::CUmemoryPool,
) -> Result<sys::CUdeviceptr, DriverError> {
    let mut dev_ptr = MaybeUninit::uninit();
    sys::cuMemAllocFromPoolAsync(dev_ptr.as_mut_ptr(), num_bytes, pool, stream).result()?;
    Ok(dev_ptr.assume_init())
}
```
 
**`CudaMemPool`** (`cudarc_shim.rs`): RAII wrapper around `CUmemoryPool`, following the same ownership pattern as `CudaStream` and `CudaContext`; constructor calls `cuMemPoolCreate`, `Drop` calls `cuMemPoolDestroy`, and the raw handle is exposed via `.handle()` for use in allocation calls.
 
**`CudaContext::default_mem_pool`**: Accessor for the device's default pool handle (via `cuDeviceGetDefaultMemPool`), useful for configuring pool attributes without creating a new pool.
 
**`api.rs`**: Added `malloc_from_pool_async` high-level wrapper matching the existing `malloc_async` signature pattern.
 
> Note: `CudaContext` already cached `has_async_alloc` with the annotation `"cached device capability for future async memory pool decisions"`. This PR is (I believe?) the feature that field was anticipating.
 
### `cuda-async` — Context-level pool routing
 
**`AsyncDeviceContext`**: Added an `Option<CUmemoryPool>` field, initialized to `None`, preserving the default allocation behavior for all existing code.
 
**`set_device_pool`**: Public setter that configures (or clears) the pool on a device context. Follows the same `with_global_device_context_mut` pattern used by kernel caching and policy configuration.
 
**`device_alloc_async`**: The central routing function. All allocation call sites are redirected here instead of calling `cuda_core::malloc_async` directly:
 
```rust
pub unsafe fn device_alloc_async(
    num_bytes: usize,
    stream: &Arc<CudaStream>,
    device_id: usize,
) -> CUdeviceptr {
    let pool = with_global_device_context(device_id, |dc| dc.pool)
        .expect("Failed to get device context for allocation.");
    match pool {
        Some(pool) => cuda_core::malloc_from_pool_async(num_bytes, stream, pool),
        None => cuda_core::malloc_async(num_bytes, stream),
    }
}
```
 
### `cutile` — Call site migration
 
Four `malloc_async` call sites in `tensor.rs` and `api.rs` are replaced with `device_alloc_async`:
 
- `Tensor::uninitialized` - the bottleneck for `zeros`, `ones`, `full`, `arange`
- `CopyDeviceToDevice::execute`
- `CopyHostToDevice::execute`
- `CopyHostVecToDevice::execute`
 
This covers every allocation path in the crate. The change at each site is mechanical, replacing `malloc_async(num_bytes, ctx.get_cuda_stream())` with `device_alloc_async(num_bytes, ctx.get_cuda_stream(), ctx.get_device_id())`.
 
---
 
## Testing
 
Tests are implemented in `cutile/tests/pool_allocation.rs` but [**cannot run until GPU CI runners are available**.](https://github.com/NVlabs/cutile-rs/issues/13#issuecomment-4084746034) The file lives in `cutile/tests/` rather than `cuda-async/tests/` because every meaningful pool test requires end-to-end validation through the tensor API (`api::zeros` → kernel write → host readback → value check), and `cutile` depends on `cuda-async` (not the reverse).

Eight tests cover distinct failure modes:

| Test | Exercises |
|---|---|
| `pool_alloc_zeros_roundtrip` | Full lifecycle: pool create → context set → `zeros` → readback → verify → teardown → pool drop |
| `pool_alloc_ones_roundtrip` | Non-zero fill via `full_apply` kernel on pool-allocated memory |
| `pool_alloc_arange_roundtrip` | Sequential write kernel (`arange_apply`) — per-index corruption check |
| `revert_to_default_after_pool_cleared` | `set_device_pool(None)` reverts cleanly; both pool-era and post-pool tensors valid |
| `multiple_allocs_from_same_pool` | Three live tensors from one pool simultaneously — pool bookkeeping under pressure |
| `pool_alloc_multidimensional` | 2D tensor shape/stride metadata unaffected by allocation path change |
| `pool_alloc_device_copy` | `CopyDeviceToDevice::execute` call site (separate from `Tensor::uninitialized`) routes through pool |
| `default_path_without_pool_unchanged` | Baseline regression — no pool ever configured, default `cuMemAllocAsync` path unbroken |

Each test runs on a fresh thread via `common::with_test_stack`, matching the repo convention for isolating thread-local `DEVICE_CONTEXTS` state. The `setup_pool` / `teardown_pool` helpers enforce correct ordering: pool outlives all tensors, context is cleared before pool destruction.
```rust
// Representative test structure (pool_alloc_zeros_roundtrip):
let pool = unsafe { setup_pool(device_id) };

let tensor = api::zeros::([256]).sync()?;
let host_vec = tensor.to_host_vec().sync()?;
assert!(host_vec.iter().all(|&v| v == 0.0));

unsafe { teardown_pool(device_id) };
drop(pool);
```

**Not tested (deferred to follow-up):**
- Cross-thread allocation/deallocation (low risk — `cuMemFreeAsync` handles pool provenance at the driver level)
- Pool attribute configuration (`cuMemPoolSetAttribute`)
- Performance benchmarking of pool vs. default allocation
---
 
## Future work (based on any feedback, discussion, etc).
 
- **Per-tensor `_in` variants**: For use cases requiring mixed pool allocation within a single execution context (e.g., scratch buffers on a fast pool, long-lived weights on a different pool). This would layer `uninitialized_in(len, pool)` / `zeros_in(shape, pool)` on top of the context-level default.
- **Pool attribute configuration**: Exposing `cuMemPoolSetAttribute` for controlling release thresholds and trimming behavior, which should be useful for tuning memory reuse in long-running training loops.
- **`ExecutionContext`-level pool**: Currently the pool is resolved via `device_id` at allocation time. An alternative is carrying the pool on `ExecutionContext` directly, which would eliminate the `with_global_device_context` lookup on every allocation. Worth benchmarking once the feature is in use.

Please let me know if there is anything else you'd like to see out of this PR @elibol!

### DIFF
--- a/cuda-async/src/device_context.rs
+++ b/cuda-async/src/device_context.rs
@@ -313,12 +313,12 @@ where
 ///
 /// Pass `None` to revert to the default `cuMemAllocAsync` path.
 ///
-/// The pool is held via `Arc`, so it cannot be desroyed while any device
+/// The pool is held via `Arc`, so it cannot be destroyed while any device
 /// context still references it. 
 /// 
 /// # Errors
 ///
-/// Returns `DeviceError:Context` if the pool's device ordinal does not
+/// Returns `DeviceError::Context` if the pool's device ordinal does not
 /// match `device_id`, preventing cross-device mismatches.
 pub fn set_device_pool(
     device_id: usize,
@@ -344,13 +344,13 @@ pub fn set_device_pool(
 /// Allocate device memory, routing through the configured pool if one is set.
 /// 
 /// This is the pool-aware replacement for calling `cuda_core::malloc_async` directly.
-/// If no pool is configured for `device_id`, falls back to cuMemAllocAsync`.
+/// If no pool is configured for `device_id`, falls back to `cuMemAllocAsync`.
 ///
 /// # Safety
 /// `stream` must be a valid CUDA stream.
 ///
 /// # Errors
-/// Returns `DeviceError` if the context lookup or allocation failes (e.g., OOM)
+/// Returns `DeviceError` if the context lookup or allocation fails (e.g., OOM)
 pub unsafe fn device_alloc_async(
     num_bytes: usize,
     stream: &Arc<cuda_core::CudaStream>,
--- a/cutile/src/tensor.rs
+++ b/cutile/src/tensor.rs
@@ -215,9 +215,10 @@ use cuda_async::device_operation::{value, DeviceOperation, ExecutionContext};
 use cuda_async::error::DeviceError;
 use cuda_core::sys::CUdeviceptr;
 use cuda_core::CudaStream;
-use std::fmt::Debug;
 use std::marker::PhantomData;
+use std::future::IntoFuture;
 use std::mem::MaybeUninit;
+use std::fmt::Debug;
 use std::ops::Index;
 use std::sync::Arc;
 
--- a/cutile/tests/pool_allocation.rs
+++ b/cutile/tests/pool_allocation.rs
@@ -19,7 +19,8 @@ use cuda_async::device_operation::DeviceOperation;
 use cuda_core::CudaContext;
 use cuda_core::CudaMemPool;
 use cutile::api;
-use cutile::tensor::ToHostVec;
+use cutile::tensor::{CopyToDevice, CopyToDeviceTensor, ToHostVec};
+use candle_core;
 use std::sync::Arc;
 
 mod common;
@@ -64,7 +65,7 @@ fn teardown_pool(device_id: usize) {
 /// - `DeviceBox::drop` → `cuMemFreeAsync` (unchanged, handles pool provenance)
 #[test]
 fn pool_alloc_zeros_roundtrip() {
-    commoon::with_test_stack(|| {
+    common::with_test_stack(|| {
         let device_id = get_default_device();
 
         // Pool must outlive tensors - declared first, dropped last.
@@ -102,7 +103,7 @@ fn pool_alloc_ones_roundtrip() {
     });
 }
 
-// Pool allocations with `api::arange` - verifies the `arange_apply` kernel path,
+/// Pool allocations with `api::arange` - verifies the `arange_apply` kernel path,
 /// which writes sequential values into pool-allocated memory.
 #[test]
 fn pool_alloc_arange_roundtrip() {
@@ -238,6 +239,78 @@ fn pool_alloc_device_copy() {
     });
 }
 
+/// Host-to-device copy via `CopyHostToDevice::execute` (Candle Tensor path).
+/// 
+/// This exercises the `device_alloc_async` call site in `CopyHostToDevice`,
+/// which is distinct from both `Tensor::uninitialized` and `CopyDeviceToDevice`.
+#[test]
+fn pool_alloc_host_to_device_copy() {
+    common::with_test_stack(|| {
+        use candle_core::{DType, Device, Tensor as CandelTensor};
+        use cutile::tensor::CopyToDevice;
+
+        let device_id = get_default_device();
+        let pool = unsafe { setup_pool(device_id) };
+
+        // Create a CPU tensor via Candle, then copy to GPU through pool.
+        let cpu_tensor = Arc::new(
+            CandleTensor::arange(0f32, 64f32, &Device::Cpu)
+                .expect("Failed to create Candle tensor."),
+        );
+
+        let gpu_tensor: Arc<cutile::tensor::Tensor<f32>> = cpu_tensor
+            .copy_to_device()
+            .sync()
+            .expect("CopyHostToDEvice failed pool.");
+
+        let host_vec = gpu_tensor
+            .to_host_vec()
+            .sync()
+            .expect("Failed to read back host-to-device copy.");
+
+        assert_eq!(host_vec.len(), 64);
+        for (i, &v) in host_vec.iter().enumerate() {
+            assert_eq!(v, i as f32, "host-to-device mismatch at index {i}");
+        }
+
+        teardown_pool(device_id);
+        drop(pool);
+    });
+}
+
+/// Host Vec to device copy via `CopyHostVecToDevice::excute`.
+/// 
+/// This exercises the `device_alloc_async` call site in `CopyHostVecToDevice`.
+#[test]
+fn pool_alloc_host_vec_to_device_copy() {
+    common::with_test_stack(|| {
+        use cutile::tensor::CopyToDeviceTensor;
+
+        let device_id = get_default_device();
+        let pool = unsafe { setup_pool(device_id) };
+
+        let host_data: Arc<Vec<f32>> = Arc::new((0..128).map(|i| i as f32).collect());
+
+        let gpu_tensor = host_data
+            .copy_to_device_tensor()
+            .sync()
+            .expect("CopyHostVecToDevice failed with pool.");
+
+        let host_vec = gpu_tensor
+            .to_host_vec()
+            .sync()
+            .expect("Failed to read back host-vec-to-device copy.");
+
+        assert_eq!(host_vec.len(), 128);
+        for (i, &v) in host_vec.iter().enumerate() {
+            assert_eq!(v, i as f32, "host-vec-to-device mismatch at index {i}");
+        }
+
+        teardown_pool(device_id);
+        drop(pool);
+    });
+}
+
 // ---------------------------------------------------------------------------
 // Tests — regression and validation
 // ---------------------------------------------------------------------------
@@ -246,7 +319,7 @@ fn pool_alloc_device_copy() {
 /// as before - this is the baseline regression test ensuring the routing
 /// change doesn't break the default.
 #[test]
-fn default_path_wihout_pool_unchanged() {
+fn default_path_without_pool_unchanged() {
     common::with_test_stack(|| {
         // No pool setup - go straight to allocation.
         let tensor = api::zeros::<f32>([1024]).sync().expect("Default allocation path failed.");
@@ -258,7 +331,48 @@ fn default_path_wihout_pool_unchanged() {
     });
 }
 
-/// Attempting to attach a pool created on device N to a difference device
+/// Pool configured on a freshly initialized device context.
+/// 
+/// Unlike other tests which may hit a lazily-initialized context,
+/// this test forces a clean context initialization before setting
+/// the pool, exercising the path where pool is attached to a 
+/// brand-new context rather than mutated onto an existing one.
+#[test]
+fn pool_on_fresh_context_init() {
+    common::with_test_stack(|| {
+        use cuda_async::device_context::init_device_contexts;
+
+        let device_id = get_default_device();
+
+        // Explicitly initialize the context map before any lazy init happens.
+        init_device_contexts(device_id, 1)
+            .expect("Failed to explicitly init device contexts.");
+
+        let ctx = CudaContext::new(device_id).expect("Failed to create CudaContext.");
+        let pool = Arc::new(unsafe {
+            CudaMemPool::new(&ctx).expect("Failed to create CudaMemPool.")
+        });
+
+        // Attach pool to the freshly initialized context.
+        set_device_pool(device_id, Some(pool.clone()))
+            .expect("Failed to set pool on fresh context.");
+
+        let tensor = api::zeros::<f32>([128])
+            .sync()
+            .expect("Failed to allocate from pool on fresh context.");
+
+        let host_vec = tensor.to_host_vec().sync().expect("readback failed.");
+        assert!(
+            host_vec.iter().all(|&v| v == 0.0),
+            "Fresh context pool allocation produced incorrect values."
+        );
+
+        teardown_pool(device_id);
+        drop(pool);
+    });
+}
+
+/// Attempting to attach a pool created on device N to a different device
 /// must return an error, not silently accepting a cross-device handle.
 #[test]
 fn set_pool_rejects_device_mismatch() {
@@ -273,7 +387,7 @@ fn set_pool_rejects_device_mismatch() {
         let pool_on_1 = Arc::new(unsafe { CudaMemPool::new(&ctx_1).expect("Failed to create pool on device 1.") });
 
         // Attaching device-1's pool to device 0 must fail.
-        let result = set_device_pool(0, Some(pool1_on_1));
+        let result = set_device_pool(0, Some(pool_on_1));
         assert!(result.is_err(), "set_device_pool should reject cross-device pool, but got Ok.");
     });
 }
--- a/cuda-core/src/api.rs
+++ b/cuda-core/src/api.rs
@@ -87,9 +87,8 @@ pub unsafe fn malloc_from_pool_async(
     num_bytes: usize,
     stream: &Arc<CudaStream>,
     pool: sys::CUmemoryPool,
-) -> sys::CUdeviceptr {
+) -> Result<sys::CUdeviceptr, crate::DriverError> {
     crate::memory::malloc_from_pool_async(stream.cu_stream(), num_bytes, pool)
-        .expect("Malloc from pool async failed.")
 }
 
 /// Asynchronously frees device memory on the given stream.


cuda-async/src/device_context.rs
```diff
@@ -88,6 +88,7 @@ pub struct AsyncDeviceContext {
     context: Arc<CudaContext>,
     deallocator_stream: Arc<CudaStream>,
     policy: Arc<GlobalSchedulingPolicy>,
+    pool: Option<cuda_core::sys::CUmemoryPool>,
```
goog00: hi ，i think there’s  some misuse risk in the caller-owned raw handle approach.
since `pool: Option<CUmemoryPool>` only stores the raw handle and doesn’t own the pool lifetime, it seems possible for the pool to be destroyed too early while later allocations/frees still rely on that handle.

cuda-core/src/cudarc_shim.rs
```diff
@@ -1106,6 +1172,20 @@ pub mod memory {
...
+    ///
+    /// # Safety
+    /// `stream` must be a valid stream handle. `pool` must be a valid memory pool handle.
+    pub unsafe fn malloc_from_pool_async(
+        stream: sys::CUstream,
+        num_bytes: usize,
+        pool: sys::CUmemoryPool,
+    ) -> Result<sys::CUdeviceptr, DriverError> {
+        let mut dev_ptr = MaybeUninit::uninit();
+        sys::cuMemAllocFromPoolAsync(dev_ptr.as_mut_ptr(), num_bytes, pool, steam).result()?;
```
goog00: hi，this looks like a typo: should this be `stream` instead of `steam`?

cuda-async/src/device_context.rs
```diff
@@ -284,6 +286,42 @@ where
...
 }
 
+/// Set the memory pool for all subsequent allocations on `device_id`.
+///
+/// Pass `None` to revert to the default async collection path (`cuMemAllocAsync`).
+/// The caller is responsible for ensuring the pool outlives all tensors allocated from it.
+///
+/// # Safety
+/// `pool` must be a valid `CUmemoryPool` handle, or `None`.
+pub unsafe fn set_device_pool(
```
goog00: hi，one thing that stood out to me here is that `AsyncDeviceContext` is stored in thread-local state.
If I’m reading this correctly, `set_device_pool(device_id, ...)` only updates the context for the current thread on that device, rather than acting like a process-wide setting. That means if a user sets the pool on thread A but the allocation later happens on thread B, thread B may not see that pool configuration.

another thing I was wondering about is the pool/device matching assumption.
`set_device_pool(...)` doesn’t seem to validate that the `CUmemoryPool` being attached actually belongs to the same device/context as the `device_id` being configured. If that assumption is left entirely to the caller, it feels like an easy place for cross-device mismatches to slip in.

cuda-core/src/cudarc_shim.rs
ghost: ah, thank you! this is what i get for burning the late night oil on larger changes 😅

fixed in https://github.com/NVlabs/cutile-rs/pull/26/commits/60c7065e7a1d59461b292b5bddfe59a14a6f5538.

cuda-async/src/device_context.rs
ghost: **on the thread-local scoping**: this is an inherent property of the existing `AsyncDeviceContext` architecture... scheduling policy, deallocator stream, and kernel cache are all per-thread already. so the pool configuration is consistent with how the system works today. that said, i agree this deserves explicit documentation on `set_device_pool` — something like "configures the pool for the calling thread's context only; other threads must call this independently." i'll add that.

**on the device mismatch**: good catch, i completely missed this! i'll add a validation in `set_device_pool` that checks the pool's device matches `device_id` and returns a `DeviceError` on mismatch. since we own `CudaMemPool`, I'll expose the device ordinal on it and compare directly so there's no need for an extra driver query.

fixed in https://github.com/NVlabs/cutile-rs/pull/26/commits/6902b8e72c103fc46dae4062b61f2467bd25c853.

cuda-async/src/device_context.rs
ghost: great catch! you're right. i took some time to review, and i think the fix is straightforward. i'll change the field to `Option<Arc<CudaMemPool>>` and update `set_device_pool` to take `Option<Arc<CudaMemPool>>` instead of a raw handle. this makes the pool's lifetime tied to the context via `Arc`, matching the existing pattern on the same struct (`context: Arc<CudaContext>`, `deallocator_stream: Arc<CudaStream>`). the pool can't be destroyed while any context still references it.

fixed in https://github.com/NVlabs/cutile-rs/pull/26/commits/09f45085cd954f82c6e463d6cf58b1bc9ab4c5a7.

cuda-async/src/device_context.rs
```diff
@@ -284,6 +286,65 @@ where
...
+                device_id,
+                &format!(
+                    "Pool belongs to device {}, but target device is {}.",
+                    p.device_ordinal(),
+                    device_id,
+                ),
+            ));
+        }
+    }
+    with_global_device_context(device_id, |dc| {
```
goog00: hi, i might be missing context, but this looks like a write through an immutable accessor. `dc.pool = pool` needs `&mut AsyncDeviceContext`, while `with_global_device_context` provides `&AsyncDeviceContext`.  
would switching to `with_global_device_context_mut` be the intended fix here?

cutile/src/api.rs
```diff
@@ -182,7 +183,7 @@ impl<T: WithDType + Send> DeviceOperation for CopyDeviceToDevice<T> {
         let num_elements = tensor.size();
         let num_bytes = element_size * num_elements;
         let src = tensor.cu_deviceptr();
-        let dst = malloc_async(num_bytes, ctx.get_cuda_stream());
+        let dst = malloc_async(num_bytes, ctx.get_cuda_stream(), ctx.get_device_id());
```
goog00: hi,this call site still seems to be using `malloc_async()` . i think this should probably route through `device_alloc_async()` as well

cutile/src/tensor.rs
```diff
@@ -212,7 +212,8 @@ use cuda_async::device_operation;
 use cuda_async::device_operation::{value, DeviceOperation};
 use cuda_async::error::DeviceError;
 use cuda_core::sys::CUdeviceptr;
-use cuda_core::{malloc_async, CudaStream};
+use cuda_core::device_context::device_alloc_async;
+use cuda_core::CudaStream,
```
goog00: this line looks like it should probably end witd `;` instead of `,`

ghost: Fixed in https://github.com/NVlabs/cutile-rs/pull/26/commits/4b02c786a480dd8f0e7f65fd237dde0e8a8e4305.

cuda-async/src/device_context.rs
ghost: Fixed in https://github.com/NVlabs/cutile-rs/pull/26/commits/fa2a63b1be82a272cfc9cc5a53b5c9301ea6de76.

cutile/src/api.rs
ghost: Fixed in https://github.com/NVlabs/cutile-rs/pull/26/commits/231c86a7ca70eff1e9958737963bd8285f7925b3.

cuda-core/src/cudarc_shim.rs
```diff
@@ -336,6 +336,79 @@ impl CudaStream {
     }
 }
 
+/// An owned CUDA memory pool, destroyed on drop.
+///
+/// Wraps a `CUmemoryPool` created via `cuMemPoolCreate`. For the device's
+/// default pool (obtained via `cuDeviceGetDefaultMemPool`), use the raw
+/// handle directly - default pools are not owned and must not be destroyed.
+#[derive(Debug)]
+pub struct CudaMemPool {
```
goog00, line 345: i may be missing some context, but it looks like `malloc_from_pool_async` and `CudaMemPool` are each added twice in cudarc_shim.rs and api.rs?

cuda-async/src/device_context.rs
```diff
@@ -148,15 +149,31 @@ pub fn init_device_contexts_default() -> Result<(), DeviceError> {
...
-/// runtime auto-initialize with the default policy.
+/// When `pool` is `Some`, all allocations via [`device_alloc_async`] on this
+/// context will route through `cuMemAllocFromPoolAsync`. The pool's device
+/// ordinal must match `device_id`.
+///
+/// Most users should use [`init_device`] or let the runtime auto-initialize.
 pub fn new_device_context(
     device_id: usize,
     mut policy: GlobalSchedulingPolicy,
+    pool: Option<Arch<CudaMemPool>>
```
goog00: Hi, this looks like a typo : `Arc` instead of `Arch`

cutile/src/api.rs
```diff
@@ -182,7 +182,7 @@ impl<T: WithDType + Send> DeviceOperation for CopyDeviceToDevice<T> {
         let num_elements = tensor.size();
         let num_bytes = element_size * num_elements;
         let src = tensor.cu_deviceptr();
-        let dst = malloc_async(num_bytes, ctx.get_cuda_stream());
+        let dst = device_alloc_async(num_bytes, ctx.get_cuda_stream(), ctx.get_device_id());
```
goog00: small thing, but `device_alloc_async` now returns a Result, while these call sites still use it like a raw pointer. Should these be updated to propagate `?` so the new allocation error path actually works?

cutile/src/api.rs
```diff
@@ -253,7 +253,7 @@ impl<T: WithDType + Send> DeviceOperation for CopyHostToDevice<T> {
         let (vec, shape, strides) = params;
         let element_size = std::mem::size_of::<T>();
         let num_elements = vec.len();
-        let dptr = malloc_async(element_size * num_elements, ctx.get_cuda_stream());
+        let dptr = device_alloc_async(element_size * num_elements, ctx.get_cuda_stream(), ctx.get_device_id());
```
goog00: same here

cutile/src/api.rs
```diff
@@ -444,7 +444,7 @@ impl<T: WithDType + Send> DeviceOperation for CopyHostVecToDevice<T> {
         let num_elements = vec.len();
         let shape = vec![num_elements as i32];
         let strides = vec![1];
-        let dptr = malloc_async(element_size * num_elements, ctx.get_cuda_stream());
+        let dptr = device_alloc_async(element_size * num_elements, ctx.get_cuda_stream(), ctx.get_device_id());
```
goog00: same here

ptxpapi:
@elibol there's no need to apologize! i can only imagine how busy you and many others are right now.

if anything, i should apologize — i've actually been making these changes off of an AMD based laptop (integrated cpu + gpu 😅) and don't have a desktop pc with a gpu (nor can i afford to procure one in a timely manner).

if anything, would it be alright if i rent a gpu from somewhere like lambda or sfcompute over the weekend? may be a quick + cheap workaround on a timeline that would allow us to validate these changes before those runners come online.

cutile/src/api.rs
ghost: Fixed in https://github.com/NVlabs/cutile-rs/pull/26/commits/e5fcefadfbe7999ecc2c2a68de32361b336de33a.

cutile/src/api.rs
ghost: Fixed in https://github.com/NVlabs/cutile-rs/pull/26/commits/e5fcefadfbe7999ecc2c2a68de32361b336de33a.

cutile/src/api.rs
ghost: you're correct, i meant to include this in the original change but completely forgot to. 

Fixed in https://github.com/NVlabs/cutile-rs/pull/26/commits/e5fcefadfbe7999ecc2c2a68de32361b336de33a.

cuda-async/src/device_context.rs
ghost: Fixed in https://github.com/NVlabs/cutile-rs/pull/26/commits/e5fcefadfbe7999ecc2c2a68de32361b336de33a.

cuda-core/src/cudarc_shim.rs
ghost: This, along with the other errors you caught, are yet again a symptom of late night contributions 😅

Apologies for all the errors! Fixed in https://github.com/NVlabs/cutile-rs/pull/26/commits/e5fcefadfbe7999ecc2c2a68de32361b336de33a.

cutile/tests/pool_allocation.rs
```diff
@@ -0,0 +1,279 @@
...
+// Helpers
+// ---------------------------------------------------------------------------
+ 
+/// Creates a `CudaMemPool` on the default device and configures the thread-local
+/// context to allocate from it. Returns the `Arc<CudaMemPool>` - the context
+/// holds a clone, so the pool stays alive as long as either reference exists.
+///
+/// # Safety
+/// Must be called from a thread with a valid CUDA context for `device_id`.
+unsafe fn setup_pool(device_id: usize) -> Arc<CudaMemPool> {
```
goog00, line 38: hi，i think there are may be a test gap here: the current coverage seems to  go through `setup_pool()` `set_device_pool` ,but not the path where pool-aware behavior is during context/device initialization. if that regresses later,the current tests may not catch it.

cutile/tests/pool_allocation.rs
```diff
@@ -0,0 +1,279 @@
...
+    })
+}
+
+/// Device-to-device copy from a pool-allocated source tensor.
+///
+/// The `copy` path uses its own `device_alloc_async` call site
+/// (`CopyDeviceToDevice::execute`), distinct from `Tensor::uninitialized`.
+/// This test ensures that path also routes through the pool correctly.
+#[test]
+fn pool_alloc_device_copy() {
```
goog00, line 223: hi, i think we're missing tests for `CopyHostToDevice` and` CopyHostVecToDevice`. If pool routing is broken  on cpu->gpu transfer, the current tests could still all pass.

cuda-core/src/cudarc_shim.rs
goog00: thanks for the quick fix! No worries, i‘ll keep reviewing the updated pr

cutile/tests/pool_allocation.rs
ghost: good point. the builder path (`new_device_context(..., pool)`) isn't directly testable from integration tests. `init_device` takes a `&mut HashMap` that's internal to the thread-local `DEVICE_CONTEXTS`, so we can't call it from outside the crate. what we can do is test pool attachment on a freshly initialized context, which is the closest exercisable path. i've added `pool_on_fresh_context_init` which calls `init_device_contexts` explicitly before `set_device_pool`, rather than relying on lazy init like the other tests. it's not a perfect proxy for the builder, but it catches regressions in the "pool on a brand-new context" flow. if we want full builder coverage, we'd need a unit test inside `cuda-async` itself — happy to add that as a follow-up (if you think it's necessary within this scope @elibol ).

Fixed in https://github.com/NVlabs/cutile-rs/pull/26/commits/5fa00f5ef6158449d2de5c037277b74c590fa42d.

cutile/tests/pool_allocation.rs
ghost: you're right... we modified four allocation call sites, but only tested three. i've gone ahead and added `pool_alloc_host_to_device_copy` (which exercises `CopyHostToDevice` via Candle's `copy_to_device`) and `pool_alloc_host_vec_to_device_copy` (which exercises `CopyHostVecToDevice` via `copy_to_device_tensor`). all four `device_alloc_async` call sites now (to my knowledge) have dedicated coverage.

Fixed in https://github.com/NVlabs/cutile-rs/pull/26/commits/5fa00f5ef6158449d2de5c037277b74c590fa42d.

PR #47, opened 2026-03-26
Cleanup cutile-compiler

ur4t:


elibol:
Thanks for the cleanup! Since this touches compiler internals (`_function.rs`, `_type.rs`, `_value.rs`, `bounds.rs`), I want to look at it more closely. We also have a compiler rewrite in progress that will likely make these changes obsolete. It might make sense to wait until that lands and then take a pass on the updated compiler (if you're interested). Will follow up.

ur4t:
> We also have a compiler rewrite in progress that will likely make these changes obsolete. It might make sense to wait until that lands and then take a pass on the updated compiler.

Sure. I will rebase when it lands.

ur4t:
I investigated functions marked `#[allow(clippy::too_many_arguments)]`. Some of them, especially `fhma`s, require those arguments natively. The others are about `compile`, from which we can indeed extract shared arguments into an argument "bundle".

I thought `#[allow(clippy::too_many_arguments)]` might be conspicuous enough so I did not add `// TODO`.

elibol:
@ur4t thanks for these. Rather than ask you to rebase across three months of churn, I am carrying the lint policy and the cleanups forward on top of current main in a follow-up PR, with you credited as co-author, and closing these two in its favor. I will link the PR here once it is up.

elibol:
Carried forward in #263 (you are credited as co-author on the squash).

PR #48, opened 2026-03-26
Cleanup cutile

ur4t:


elibol (approved):
Same note about `#[allow(clippy::too_many_arguments)]` here too -- otherwise LGTM!

PR #49, opened 2026-03-26
Cleanup examples and benchmarks

ur4t:


elibol (approved):
LGTM! The iterator improvements and clone removals look good.

Minor suggestion (for next time): for the `#[allow(clippy::too_many_arguments)]` additions, it might be worth leaving a `// TODO` so we remember to revisit these with config structs or similar later.

Thanks for the cleanup!

PR #50, opened 2026-03-26
feat: Support zero-copy Arc-backed view ,flatten, and reinterpret

goog00:
## Summary

This PR adds zero-copy tensor view/flatten/reinterpret support, including `Arc`-backed immutable views. The main goal is to allow rank-changing views without cloning or device copies, while keeping them compatible with the existing kernel launcher paths.

Related to #6 

## Design

The main change is that `Tensor<T>` now uses shared backing storage plus per-view metadata (`shape`, `strides`, and logical element type `T`), so multiple tensor views can share the same device allocation.

For this first pass, I intentionally kept the implementation focused on contiguous tensors only. It also adds the necessary byte-size, alignment, and storage-uniqueness checks for the new view/reinterpret paths.

## Tests

I added coverage for:
- zero-copy pointer identity
- invalid shape / byte-size rejection
- misaligned storage rejection for reinterpret
- mutable partition rejection for shared storage
- reinterpret dtype coverage
- kernel interop for view / reinterpret paths

All existing and new tests pass locally.

## Follow-up

This does not yet cover general non-contiguous tensor views, storage offsets, or layout-preserving reinterpret for arbitrary strides. If broader view semantics are desirable, those would make sense as follow-up work.

I tried to keep this initial version fairly small and focused on the contiguous case.
If this seems like a reasonable scope for a first pass, I’d be very happy to hear any feedback. And if you were hoping for a different subset or API direction first, I’d really appreciate your thoughts.

### DIFF
--- a/cuda-async/src/device_box.rs
+++ b/cuda-async/src/device_box.rs
@@ -37,53 +37,56 @@ impl<T> DevicePointer<T> {
     }
 }
 
-/// An owning handle to a CUDA device memory allocation, freed asynchronously on drop.
+/// An owning, type-erased handle to a CUDA device memory allocation, freed asynchronously on drop.
+///
+/// `DeviceBox` manages a raw byte buffer on the GPU. It stores only the device pointer,
+/// byte length, and device id — all type information lives in higher-level wrappers
+/// such as `Tensor<T>`.
 #[derive(Debug)]
-pub struct DeviceBox<T: Send + ?Sized> {
-    _device_id: usize,
-    dtype: PhantomData<T>,
+pub struct DeviceBox {
+    device_id: usize,
     cudptr: CUdeviceptr,
     len: usize,
 }
 
-unsafe impl<T: Send + ?Sized> Send for DeviceBox<T> {}
-unsafe impl<T: Send + ?Sized> Sync for DeviceBox<T> {}
+unsafe impl Send for DeviceBox {}
+unsafe impl Sync for DeviceBox {}
 
-impl<T: Send + ?Sized> Drop for DeviceBox<T> {
+impl Drop for DeviceBox {
     fn drop(&mut self) {
         unsafe {
             // Safety: The CUDA driver is guaranteed to complete any queued async operations.
-            with_deallocator_stream(self._device_id, |stream| {
+            with_deallocator_stream(self.device_id, |stream| {
                 free_async(self.cudptr, stream);
             })
             .unwrap_or_else(|_| {
                 panic!(
                     "Failed to free device pointer on device_id={}",
-                    self._device_id
+                    self.device_id
                 )
             })
         }
     }
 }
 
-impl<DType: Send + Sized> DeviceBox<[DType]> {
-    /// Constructs a `DeviceBox<[DType]>` from a raw device pointer, length, and device id.
+impl DeviceBox {
+    /// Constructs a `DeviceBox` from a raw device pointer, byte length, and device id.
     ///
     /// # Safety
-    /// The caller must ensure `dptr` points to a valid device allocation of at least `len` elements.
-    pub unsafe fn from_raw_parts(dptr: CUdeviceptr, len: usize, device_id: usize) -> Self {
+    /// The caller must ensure `dptr` points to a valid device allocation of at least
+    /// `len_bytes` bytes.
+    pub unsafe fn from_raw_parts(dptr: CUdeviceptr, len_bytes: usize, device_id: usize) -> Self {
         Self {
-            dtype: PhantomData,
             cudptr: dptr,
-            len,
-            _device_id: device_id,
+            len: len_bytes,
+            device_id,
         }
     }
-    /// Returns whether there is no element in the device slice.
+    /// Returns whether the allocation is empty (zero bytes).
     pub fn is_empty(&self) -> bool {
         self.len == 0
     }
-    /// Returns the number of elements in the device slice.
+    /// Returns the byte length of the allocation.
     pub fn len(&self) -> usize {
         self.len
     }
@@ -93,13 +96,6 @@ impl<DType: Send + Sized> DeviceBox<[DType]> {
     }
     /// Returns the device id this allocation belongs to.
     pub fn device_id(&self) -> usize {
-        self._device_id
-    }
-    /// Returns a non-owning `DevicePointer` to the underlying allocation.
-    pub fn device_pointer(&self) -> DevicePointer<DType> {
-        DevicePointer::<DType> {
-            dtype: PhantomData,
-            dptr: self.cudptr,
-        }
+        self.device_id
     }
 }
--- a/cutile/src/tensor.rs
+++ b/cutile/src/tensor.rs
@@ -390,8 +390,9 @@ pub trait IntoPartitionArc {
 ///
 /// ## Memory Management
 ///
-/// Tensors own their GPU memory through a `DeviceBox`. Memory is automatically freed
-/// when the tensor is dropped. For shared ownership, use `Arc<Tensor<T>>`.
+/// Tensors share GPU memory ownership through `Arc<DeviceBox>`. Memory is automatically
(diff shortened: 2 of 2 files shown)


cutile/src/api.rs
```diff
@@ -184,12 +183,13 @@ impl<T: WithDType + Send> DeviceOperation for CopyDeviceToDevice<T> {
...
-            shape: shape.clone(),
-            strides: strides.clone(),
-        })
+        Ok(Tensor::from_raw_parts(
+            dst,
+            num_bytes,
+            ctx.get_device_id(),
+            shape.clone(),
+            strides.clone(),
+        ))
```
elibol, line 191: Can you help explain why you prefer this interface over device_box?

cutile/src/tensor.rs
```diff
@@ -429,14 +444,171 @@ pub trait IntoPartitionArc {
...
 /// ```
+#[derive(Debug)]
+pub(crate) struct TensorStorage {
+    device_box: DeviceBox<[u8]>,
+}
+
 #[derive(Debug)]
 pub struct Tensor<T: WithDType> {
-    pub device_box: DeviceBox<[T]>,
+    pub(crate) storage: Arc<TensorStorage>,
```
elibol: This is the pattern I had in mind, but I was thinking this would be a distinct type (like TensorView). In particular, I was thinking of getting rid of `Arc<Tensor<T>>` in favor of a simple `TensorView<T>`. Thoughts?

That said, I like this approach because it's not too invasive and gets the behavior we're looking for. I can look into the DeviceOperation API, once this lands. I just need help understanding these key things to make sure we're not introducing any safety leaks:

1. We cannot construct a view from a plain `Tensor<T>`. It needs to be `Arc<Tensor<T>>`.
2. When the last `Arc<Tensor<T>>` goes out of scope, `Tensor<T>` goes out of scope, and the internal `Arc<TensorStorage>` goes out of scope, triggering the drop impl for DeviceBox.
3. The usual constraint that `Partition<Tensor<T>>` is needed for mutable access to `Tensor<T>` from within a kernel still applies.

Have I understood the impl correctly?

If so, the only thing I would say is that it seems like the TensorStorage impl is working around the DeviceBox impl. The type interface for DeviceBox seems redundant. If the type-erased version is preferred / more idiomatic for this pattern (views), then let's just do that and get rid of DeviceBox.

goog00: @elibol Thanks for the detailed review — this is very helpful.

On `TensorView` vs `Arc<Tensor<T>>`: that makes sense, and I can see the appeal of a distinct view type. I went with the current approach because it felt like the least invasive way to get the zero-copy behavior working with the existing launcher and host-side tensor APIs. But I agree that a dedicated `TensorView<T>` could make the model clearer. If you think that would be the better long-term direction, I’d be happy to rework this toward a `TensorView<T>`-based design.

Yes, I believe your understanding of the current implementation is correct on all three points:

- In the current implementation, zero-copy views are only created from `&Arc<Tensor<T>>`. A plain `Tensor<T>` can still `reshape` / `flatten` by consuming itself, but shared zero-copy views currently go through the `Arc` path.

- The drop chain is: last `Arc<Tensor<T>>` drops -> `Tensor<T>` drops -> last `Arc<TensorStorage>` drops -> `DeviceBox<[u8]>` drops -> `free_async`. Shared views keep the underlying allocation alive through the inner `Arc<TensorStorage>`.

- Mutable kernel access still goes through `Partition<Tensor<T>>`. `partition()` takes `self` by value and checks `assert_unique_storage()`, so if any view is still sharing the same storage, mutable partitioning is rejected. That is what the `shared_storage_blocks_mutable_partition` test is intended to validate.


On `TensorStorage` vs `DeviceBox`: I think that is a very fair point. `TensorStorage` is currently acting as a thin workaround for the still-typed `DeviceBox<[T]>` interface, while the storage model here is now effectively type-erased. I agree that pushing that type erasure down into `DeviceBox` would likely be the cleaner design. I’m happy to revise this PR in that direction rather than keeping `TensorStorage` as an extra layer.

cutile/src/api.rs
goog00: With the zero-copy view model, Tensor now owns its storage through `Arc<TensorStorage> `rather than holding a `DeviceBox<[T]> `directly. `Tensor::from_raw_parts` encapsulates this — it creates the DeviceBox<[u8]>, wraps it in TensorStorage, and validates shape/strides against the byte length, all in one place.

Previously each call site in api.rs was assembling the `DeviceBox` and Tensor struct fields individually, which meant the type-erasure logic ([u8] instead of [T]) and validation were scattered across multiple locations. This consolidates that into a single constructor.

That said, as we discussed below, if we simplify `DeviceBox` to be untyped, this extra layer becomes less necessary.

cutile/src/tensor.rs
elibol: I think it's fine the way it is. We can revisit the TensorView / TensorStorage topics at a later time.

If you don't mind, please resolve the conflicts. I would but it's kind of hard to do through the github interface :)

I think the main conflicts are with the WithDType -> DType refactor.

goog00: Sounds good! I'll rebase and resolve the conflicts.

goog00: hi，I‘ve tried to refine the tensor memory model in this pr based on your comments. `DeviceBox` is now a type-erased byte buffer instead of a generic DeviceBox<T> ,and i've removed `TensorStorage` entirely. This simplifies the storage hierarchy from three layers to two without changing any zero-copy view semantics or safety invariants.

I’ve also rebased and resolved the conflicts in this PR.

All tests are passing,let me know if there's anything i should  adjust.

cuda-async/src/device_box.rs
```diff
@@ -24,55 +24,69 @@ impl<T> DevicePointer<T> {
...
+/// byte length, and device id — all type information lives in higher-level wrappers
+/// such as `Tensor<T>`.
 #[derive(Debug)]
-pub struct DeviceBox<T: Send + ?Sized> {
-    _device_id: usize,
-    dtype: PhantomData<T>,
+pub struct DeviceBox {
+    device_id: usize,
     cudptr: CUdeviceptr,
     len: usize,
```
elibol, line 49: This is pretty high priority so I'm not going to block on this, but we should rename this to `len_bytes`.

cuda-async/src/device_box.rs
```diff
@@ -24,55 +24,69 @@ impl<T> DevicePointer<T> {
...
         }
     }
-    /// Returns whether there is no element in the device slice.
+    /// Returns whether the allocation is empty (zero bytes).
     pub fn is_empty(&self) -> bool {
         self.len == 0
     }
-    /// Returns the number of elements in the device slice.
+    /// Returns the byte length of the allocation.
     pub fn len(&self) -> usize {
```
elibol, line 90: This should also be `len_bytes`.

cutile/src/tensor.rs
```diff
@@ -429,12 +430,164 @@ pub trait IntoPartitionArc {
...
+        )
+    }
+
+    // Returns the physical byte length of the shared backing allocation.
+    fn storage_num_bytes(&self) -> usize {
+        self.storage.len()
+    }
+
+    // Returns the logical element count described by the tensor's shape metadata.
+    fn logical_num_elements(&self) -> usize {
```
elibol, line 540: nit: rename to `num_elements`.

cutile/src/tensor.rs
```diff
@@ -429,12 +430,164 @@ pub trait IntoPartitionArc {
...
+        checked_num_elements_i32(&self.shape)
+            .expect("Tensor shape contains invalid dimensions or overflows.")
+    }
+
+    // Returns the logical byte size implied by shape metadata and dtype T.
+    fn logical_num_bytes(&self) -> usize {
```
elibol, line 546: nit: rename to `num_bytes`.

PR #55, opened 2026-03-27
Add interoperability guide for custom CUDA kernels

### LINKED ISSUE #51: Warp primitives
Hi! 
At work we have quite a bit of Rust tooling related to running CUDA kernels so this project is very interesting to me.
Question: I haven't worked with the tile model of kernels. I read the book you released and it looks great.
However: At work we have some kernels which could only get fast enough by using warp primitives/functions: https://docs.nvidia.com/cuda/cuda-programming-guide/05-appendices/cpp-language-extensions.html#warp-functions

If we assume we're using cutile-rs for 95% of kernels but some need special care via things like warp functions, what's the recommended approach?

I suppose one "escape hatch" is always to sync, use device pointers, drop into C++ (via our current Rust stack), launch the fast kernel, then a similar path to come back to cutile-rs.
What I wonder is if there are other possibilities now or if you have some plans down the line to support something like this.

Thanks!

elibol:
## Summary
- New guide chapter: **Interoperability with Custom CUDA Kernels**
- Documents when tile programming is the right fit vs. when warp-level primitives are needed
- Shows how to load and launch custom CUDA C++ kernels via `AsyncKernelLaunch`
- Demonstrates composing custom kernels with tile kernels using `and_then()` and `with_context`
- Added to the User Guide table of contents between Async Execution and Performance Tuning

Resolves #51

## Test plan
- [ ] Verify the book builds: `cd cutile-book && make livehtml`
- [ ] Review code examples for accuracy against current API

### DIFF
--- a/cutile-book/guide/interoperability.md
+++ b/cutile-book/guide/interoperability.md
@@ -1,0 +1,251 @@
+# Interoperability with Custom CUDA Kernels
+
+The tile model handles dense tensor algebra well — GEMM, element-wise operations, reductions, convolutions — but some algorithms depend on **warp-level primitives** (`__shfl_sync`, `__ballot_sync`, `__reduce_sync`) for things like custom scan/prefix-sum, cooperative groups, or irregular data access patterns. For these, write the kernel in CUDA C++ and integrate it using the approach below.
+
+A custom CUDA kernel can participate in the same `DeviceOperation` execution model as your tile kernels — sharing streams, chaining with `.and_then()`, and avoiding unnecessary synchronization.
+
+## Step 1: Compile Your CUDA Kernel
+
+Compile your CUDA C++ kernel to PTX (portable) or a `.cubin` (architecture-specific):
+
+```bash
+# PTX — portable across GPU architectures, JIT-compiled at load time.
+nvcc -ptx -arch=compute_80 my_kernel.cu -o my_kernel.ptx
+
+# cubin — pre-compiled for a single architecture, no JIT overhead.
+nvcc -cubin -arch=sm_80 my_kernel.cu -o my_kernel.cubin
+```
+
+> **Architecture portability:** A `.cubin` file only runs on the exact SM architecture it was compiled for. Code compiled with `-arch=sm_80` will not load on an `sm_100` GPU. PTX avoids this problem — the CUDA driver JIT-compiles it for the target GPU at load time, at the cost of a one-time compilation delay. Prefer PTX unless you need to eliminate JIT overhead. If you must ship `.cubin` files, compile for each target architecture:
+>
+> ```bash
+> nvcc -cubin -gencode arch=compute_80,code=sm_80 \
+>              -gencode arch=compute_100,code=sm_100 \
+>              my_kernel.cu -o my_kernel.cubin
+> ```
+
+## Step 2: Load the Module and Function
+
+Use `cuda-async`'s module loading functions to load the compiled kernel:
+
+```rust
+use cuda_async::device_context::load_module_from_file;
+
+let module = load_module_from_file("my_kernel.cubin", device_id)?;
+let function = Arc::new(module.load_function("my_kernel_entry")?);
+```
+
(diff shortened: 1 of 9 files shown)


cutile-book/guide/interoperability.md
```diff
@@ -0,0 +1,361 @@
...
+- **Reductions**: sum, softmax, layer normalization
+- **Stencil computations**: convolutions, pooling
+
+For these patterns, the tile compiler often produces code competitive with hand-tuned CUDA C++, with far less effort.
+
+## When You Need Something Else
+
+Certain algorithms depend on **warp-level primitives** — operations like `__shfl_sync`, `__ballot_sync`, or `__reduce_sync` that coordinate threads within a 32-thread warp. These include:
+
+- **Warp-specialized kernels**: algorithms where different warps play different roles (e.g., producer/consumer patterns in some FlashAttention variants)
```
binarybana: The Tile compiler still generates warp specialized kernels, even if the input is not explicitly warp specialized. This is critical for flash attention performance for example (which CUDA Tile does support and optimizes for).

So you might want to clarify this point that *User-explicit* warp specialized patterns aren't a good fit for tile. And either remove or clarify that good flash attention performance for tile doesn't assume user explicit specialization, even though that is often required for good performance with raw CUDA C++.

cutile-book/guide/interoperability.md
```diff
@@ -0,0 +1,361 @@
...
+}
+launcher
+    .push_arg(num_elements as u32)
+    .set_launch_config(cfg);
+
+// .and_then() ensures the custom kernel runs on the same stream,
+// seeing all writes from the tile kernel.
+let result = value(z)
+    .and_then(move |z| {
+        launcher.and_then(move |_| value(z))
```
binarybana: This is probably my naivete on Rust Tile programming, but if I'm already `await()`ing the `y.unpartition()` future on line 189, then is the `and_then` here redundant in terms of sequencing the `custom_function` call after the `stage1_op` call?

It's also confusing to me why there are 2 `and_then` calls here. And what the `value` function does (but I haven't read the previous documentation, so I'm probably just missing context).

Providing the above feedback of a naive user in case it's helpful, feel free to ignore.

cutile-book/guide/interoperability.md
```diff
@@ -0,0 +1,361 @@
...
+op.await?;
+```
+
+This pattern gives you full access to the CUDA driver API while still participating in the `DeviceOperation` execution model.
+
+---
+
+## Coming from Triton: Warp Specialization in the Tile Model
+
+[Triton](https://triton-lang.org/) and cuTile Rust share a similar programming model — both let you write kernels in terms of tile-level operations rather than individual threads. If you are coming from Triton and rely on warp specialization (e.g., `tl.async_task` for producer/consumer patterns), this section explains what the cuTile Rust compiler handles automatically and where the models differ.
```
binarybana: Fun story: Upstream Triton doesn't have a user explicit warp specialization capability. The `tl.async_task` you describe here is only part of the "tlx" extensions that Meta added and supports in [their Triton fork](https://github.com/facebookexperimental/triton).

So you should at least clarify that here. Or simplify to remove any mention that Triton allows user explicit warp spec.

The part that a user DOES have to opt-in to when using upstream Triton is the `warp_specialize` boolean flag in the `tl.range` op. Example [here](https://github.com/triton-lang/triton/blob/3e15c1a3e3a931578d2bfc9cb06ad12628adae2c/python/tutorials/06-fused-attention.py#L69).

elibol: Right on and will do. Thanks for the feedback!

cutile-book/guide/interoperability.md
elibol: This might just be a code snippet we over engineered. I'll fix it.

cutile-book/guide/interoperability.md
elibol: I'm going to update based on your feedback.

PR #58, opened 2026-03-31
Add unsafe device pointer support for AsyncKernelLaunch + DType Refactor

elibol:
This adds support for pushing an unsafe device pointer to AsyncKernelLaunch. There are various other changes to the AsyncKernelLaunch API that are needed here to ensure the API does not have any safety bugs when used directly.

This also refactors the code to use DType instead of WithDType.


cutile/src/api.rs
```diff
@@ -482,18 +343,6 @@ pub fn copy_host_vec_to_device<T: WithDType>(
 ///
 /// A tuple of:
 /// - `Vec<T>` - Flattened tensor data
```
ur4t: These comments should be removed as well.

PR #59, opened 2026-04-01
Refine cutile-macro error

ur4t:


elibol (commented):
Thanks for cleaning this up — the error API simplification and `ok_or_else` patterns look good.

One question: the change from `.err("msg")` (one step, returns `Result`) to `.error("msg").into()` (two steps) is more verbose at each call site. Was the motivation to make `Error` usable outside of `Result` contexts (e.g., error collection), or is there another reason for the split? Wondering if there's a way to keep the one-step ergonomics while still getting the benefits of the refactor.

Otherwise LGTM!

ur4t:
> One question: the change from `.err("msg")` (one step, returns `Result`) to `.error("msg").into()` (two steps) is more verbose at each call site. Was the motivation to make `Error` usable outside of `Result` contexts (e.g., error collection), or is there another reason for the split? Wondering if there's a way to keep the one-step ergonomics while still getting the benefits of the refactor.

I started by replacing `Err(syn_err(span, "msg")` patterns and found an additional `.error("msg")` which produces `Error` is required for `ok_or_else()` and thought `.err("msg")` is a bit redundant.

Personally I perfer unified `.error()` and using `.into()` to produce `Result`s, but I do not resist to bring `.err("msg")` back.

PR #89, opened 2026-04-13
Fix divisibility units, nested mutation, if/else expressions, FTZ ops, const latency

elibol:
## Summary

- Fix stride divisibility to use elements (not bytes), matching NVT and cutile-python conventions. Document element vs byte units on DivHint/SpecializationBits.
- Fix collect_mutated_variables to recurse into nested control flow (for/while/if/loop).
- Add FTZ variants for all Tile IR ops that support flush_to_zero: addf, subf, mulf, divf, fma, rsqrt, sqrt.
- Fix if/else as tile expression: condition was compiled with the expression's return type context.
- Support const generic latency hints in load/store ops (load_ptr_tko, store_ptr_tko, load_view_tko, store_view_tko).


elibol:
Superseded by #90, which includes all changes from this branch plus the static op params work.

PR #92, opened 2026-04-14
test: Add non-divisible shape tests for 2D copies and boundary sizes.

elibol:


elibol:
Subsumed by #93 (tile-ir rewrite).

PR #93, opened 2026-04-14
rewrite: remove LLVM/MLIR dependency by adding pure Rust compiler via cutile-ir

elibol:
This PR adds `cutile-ir`, a pure Rust IR builder and bytecode writer replacing `cuda-tile-rs`/`melior`/`mlir-sys`. No `LLVM` install required.


ur4t:
Genius work! You made it!

PR #96, opened 2026-04-15
feat: Add custom memory pool support for tensor allocation

goog00:
### Summary
* Pool primitives: `MemPool` RAII wrapper in `cuda-core` with create/default/threshold/destroy lifecycle
* Device-level configuration: `set_device_pool` / `get_device_pool` / `clear_device_pool` on `AsyncDeviceContext`; pool is frozen into `ExecutionContext` at scheduling time so in-flight ops are unaffected by later changes
* Unified allocation path: `ExecutionContext::alloc_async()` routes through `cuMemAllocFromPoolAsync` when a pool is set, falls back to `cuMemAllocAsync` otherwise. All tensor/copy callsites updated
* Safety: `set_device_pool` validates device affinity; `.schedule()` now reads pool (was always `None`), aligned with `.sync()` / `.await`
* Tests: 12 integration tests covering pool lifecycle, allocation routing, freeze semantics, and the explicit `.schedule()` path; 

No behavior change when no pool is configured. 

### Test plan 
* `cargo test -p cuda-async --test pool_allocation`   passes
* `./scripts/run_all.sh` passes


goog00:
I've refactored the implementation since the last push — cleaner and less intrusive.
Happy to discuss any concerns.

@elibol  PTAL when you're available, thank you!

elibol:
No immediate concerns! We already reviewed the approach extensively. I will begin looking at these once we have v0.0.2 cut. I'd like to have your PRs merged in after that, so we can iterate and address bugs while folks try out v0.0.2. Hoping to get it out soon. We might just cut v0.1.0 immediately after v0.0.2. Just trying to align versions and features in a manageable way.

I'll prioritize this and your other PR next week :slightly_smiling_face:

goog00:
Sounds good to me, thank you for the update.

elibol:
LGTM! Merging!

cuda-core/src/runtime.rs
```diff
@@ -198,6 +198,90 @@ impl Device {
...
+    /// The returned pool is owned — it will be destroyed when the last `Arc`
+    /// is dropped. Pair with [`MemPool::set_release_threshold`] to control
+    /// when the pool returns memory to the OS.
+    pub fn new_mem_pool(self: &Arc<Self>) -> Result<Arc<MemPool>, DriverError> {
+        self.bind_to_thread()?;
+        let mut props: cuda_bindings::CUmemPoolProps = unsafe { std::mem::zeroed() };
+        props.allocType = cuda_bindings::CUmemAllocationType_enum_CU_MEM_ALLOCATION_TYPE_PINNED;
+        props.handleTypes = cuda_bindings::CUmemAllocationHandleType_enum_CU_MEM_HANDLE_TYPE_NONE;
+        props.location.type_ = cuda_bindings::CUmemLocationType_enum_CU_MEM_LOCATION_TYPE_DEVICE;
+        props.location.__bindgen_anon_1.id = self.ordinal as c_int;
```
ur4t, line 213: @goog00 @elibol `id` in `CUmemLocation_st` is actually an `int` field which should be accessed directly, but `bindgen` sometimes produces a `c_int` as expected and sometimes produces an union wrapper struct. Investigating.

ur4t: @goog00 @elibol `id` in `CUmemLocation_st` is an union wrapper in CUDA 13.2. I have multiple CUDA environments thus generated bindings vary accrodingly.

goog00: > I have multiple CUDA environments thus generated bindings vary accrodingly.

@ur4t Thanks for the investigation!  Could you share which CUDA version you're using? including pre-13.x?

ur4t: > Could you share which CUDA version you're using? including pre-13.x?

@goog00 CUDA 13.0 installed along with PyTorch. And I have checked `cuda.h` in CUDA 13.1, it uses `int`. Only CUDA 13.2 uses a union wrapper.

goog00: @ur4t Thank you for checking! I'll open a follow-up PR to handle the version compatibility across different CUDA releases.

goog00: @ur4t @elibol Fixed in #128

PR #103, opened 2026-04-19
Refactor: Document libclang prerequisite and add error message

tuananhlfc:
bindgen loads libclang at build time and needs clang's compiler-builtin
headers (stddef.h, stdarg.h, ...). On Debian/Ubuntu these ship in
libclang-common-<N>-dev, which is not a dependency of libclang*. When
they are missing, the build script panicked via .unwrap() with an
opaque ClangDiagnostic, giving users no hint at the real cause.

- build.rs: propagate the bindgen error and, on the missing-builtin-
  headers case, print an install hint (apt package + BINDGEN_EXTRA_
  CLANG_ARGS fallback) instead of panicking.
- README: document the libclang-common-<N>-dev / clang prerequisite
  under Setup.


elibol:
Thanks for the fix!

One optional refinement: the `build.rs` hint hardcodes `libclang-common-18-dev`, but users with a different libclang will install the wrong package. Something like `libclang-common-<N>-dev` with a note to match their installed version might be worth looking into.

A short non-Ubuntu hint (Fedora: `clang-devel`, Arch: `clang`, macOS: `xcode-select --install`) would help folks on other platforms, but that's a nice-to-have.

Happy to merge as-is and iterate in a follow-up, or update this PR now. Your call.

PR #106, opened 2026-04-21
fix: update GPU test script

ur4t:


scripts/run_gpu_tests.sh
```diff
@@ -14,8 +14,8 @@ run_step \
     cargo test -p cutile --test error_quality
 
 run_step \
-    "cutile-compiler GPU runtime tests" \
-    cargo test -p cutile-compiler --test gpu
+    "cutile GPU runtime tests" \
+    cargo test -p cutile --test gpu
```
elibol: This command is executed below (see line 44). Think we should rename it?

ur4t: Sorry, I have not noticed, now removed as in #101.

elibol:
@ur4t, I think we'll need to replace the stale `cargo test -p cutile-compiler --test gpu` step with `cargo test -p cutile-compiler --test compiler2_e2e` instead of removing cutile-compiler GPU/tileiras coverage entirely.

ur4t:
Subsumed by #161.

PR #128, opened 2026-04-30
fix: support CUmemLocation layout across CUDA versions

goog00:
**Summary**

CUDA 13.2 wrapped `CUmemLocation::id` in an anonymous union, so bindgen emits it as `__bindgen_anon_1.id`; CUDA 13.0/13.1 toolkits keep it as a plain `int`. 

- `cuda-bindings/build.rs`  parses the bindgen-generated types.rs with syn and sets `cu_mem_location_anon_union` cfg when the union form is detected.
- `cuda-bindings` exposes a `set_mem_location_id` helper that picks the correct access path via cfg, hiding the layout difference from callers.

- `cuda-core::Device::new_mem_pool` now goes through the helper instead of writing `__bindgen_anon_1.id `directly.

**Test Plan**
`cargo test -p cuda-async --test pool_allocation`  passes（CUDA 13.1 and CUDA 13.2）


elibol:
LGTM! Thanks for addressing this so quickly.

PR #132, opened 2026-05-06
Cleanup zip tuples

ur4t:


ur4t:
@elibol Congratulations on enabling all Clippy lints. I've been a bit busy for a while thus not responding, sorry. 

This PR (Cleanup zip tuples) itself is a "nice-to-have" improvement which also demonstrates how to leverage Rust's functional programming features. Although less straightforward, I’ve verified that the output remains the same.

PR #134, opened 2026-05-07
fix: resolve pool outside with_default_device_policy closure

### LINKED ISSUE #133: [BUG]: Custom pool silently ignored when using .sync() / .await / .into_future()
### Operating System

Ub22.04

### GPU and GPU Architecture

RTX4090

### Error Description

Calling `set_device_pool()` has no effect when allocating via `.sync()`,`.await`, or `.into_future()`. The custom pool is registered but `ExecutionContext` always sees `pool: None`, causing allocations to fall back to the default allocator.

### Cause
`ExecutionContext::new()` calls `pool_for_stream()` → `get_device_pool()` → `with_global_device_context()`. But `with_default_device_policy()` (called by `sync()`/`into_future()`) already holds the `Cell`'s value via `take()` when `ExecutionContext::new()` is invoked inside its closure. The re-entrant `take()` sees `None`, silently re-initializes the context with `pool = None`, and the registered pool is lost.

### Minimal Reproduction Example

```rust
set_device_pool(0, pool_x)?;
let op = with_context(|ctx| {
    // ctx.get_pool() is None here — pool_x is silently ignored
    value(())
});
op.sync()?;
```

goog00:
Fixes #133

Extract stream selection out of the `with_default_device_policy` closure so `with_global_device_context` releases its borrow before `ExecutionContext::new` is called.

```rust
// Before
with_default_device_policy(|policy| {
    let stream = policy.next_stream()?;
    ExecutionContext::new(stream)   // re-enters Cell → pool lost
})

// After
let stream = with_default_device_policy(|policy| policy.next_stream())??;
ExecutionContext::new(stream)       // Cell already released → pool correct
```
Applied to sync(), graph(), and all into_future() implementations.

### Tests

`cargo test -p cuda-async --test pool_allocation alloc_with_custom_pool_via_device_op` passes
`cargo test -p cuda-async --test pool_allocation alloc_with_custom_pool_via_into_future` passes
`./scripts/run_all.sh` passes

### DIFF
--- a/cuda-async/src/device_operation.rs
+++ b/cuda-async/src/device_operation.rs
@@ -331,10 +331,8 @@ pub trait DeviceOp:
     fn graph(
         self,
     ) -> Result<crate::cuda_graph::CudaGraph<<Self as DeviceOp>::Output>, DeviceError> {
-        with_default_device_policy(|policy| {
-            let stream = policy.next_stream()?;
-            self.graph_on(stream)
-        })?
+        let stream = with_default_device_policy(|policy| policy.next_stream())??;
+        self.graph_on(stream)
     }
     /// Capture this operation into a replayable [`CudaGraph`](crate::cuda_graph::CudaGraph)
     /// on an **explicit stream**.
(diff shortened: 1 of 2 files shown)


elibol:
This fix looks reasonable to me for the immediate pool issue.

One follow-up thought: I think this exposed a small design oversight in `with_global_device_context`. Since it temporarily takes the thread-local context map, `None` currently means both “not initialized yet” and “already checked out.” Longer term, it may be worth making that state explicit, e.g. `Uninitialized | Available(map) | Borrowed`, so accidental nested access gives a clear error instead of silently rebuilding context.

I don’t think we need to block this PR on that, though. Thoughts?

goog00:
Opened #141  to track the explicit enum redesign and the open design questions (Err vs panic, RAII vs poisoning). 
Happy to discuss there.

Issue #141, opened 2026-05-10
Make with_global_device_context state explicit (Uninitialized | Available | Borrowed)


goog00:
Follow-up to #133 / #134.

## Background

#134 fixes the immediate pool re-entrancy by extracting stream selection out of the closure. The underlying `Cell<Option<HashMap>>` still has the ambiguity that motivated the bug: `None` means both "uninitialized" and "already taken." As suggested in #134's review, an explicit state enum would surface re-entry as a clear error instead of silently rebuilding.

## Open design questions

1. **Re-entry behavior**: Should re-entry return `Err(DeviceError)` or panic? 
    Leaning toward panic since it's a programming error rather than a recoverable runtime condition.

2. **Panic recovery**: When the closure panics mid-execution, should the  state auto-restore to `Available` (via RAII guard) or stay `Borrowed`  to signal "potentially inconsistent state" (Mutex-style poisoning)?
   Leaning toward auto-restore since most closures here are read-only and  the borrowed map isn't mutated. But happy to go with poisoning for  `_mut` if you'd rather preserve "panic = unsafe to continue" semantics.

Will follow your call on both.

PR #144, opened 2026-05-15
refactor: surface re-entry with explicit ContextState enum

goog00:
## Summary

Addresses #141 (state ambiguity in `with_global_device_context`) and tightens related design points around the new state machine: poison granularity, recovery path, and scheduling consistency.

After this PR:
- Re-entrant `with_global_device_context*` access fails loudly instead of silently rebuilding the per-thread map.
- A panicking `_mut` callback poisons only the affected device, not every device on the thread.
- Callers get an explicit recovery path instead of a permanently-dead thread.
- `DeviceBuffer::drop` no longer aborts the process during unwind.
- Stream/pool scheduling is collapsed into a single entry point that reads both under one map borrow.

Changes:
- Replace `Cell<Option<HashMap>>` with a 3-state `ContextState` enum (`Uninitialized / Available / Borrowed`) so re-entrant access panics.
- Introduce an RAII `ContextGuard` that always restores `Available` on drop.
- Add a per-device `poisoned` flag, set only when a `_mut` callback unwinds. Read-only callback panics no longer poison.
- Add public recovery API: `is_device_poisoned`, `clear_device_poison` (keeps pool/cache), `reset_device` (drops entry, lazy rebuild).
- Make `is_device_poisoned` / `clear_device_poison` / `reset_device` pure no-ops on an uninitialized thread, matching their docs.
- Make deallocation poison-immune: `with_deallocator_stream` skips the poison check; `DeviceBuffer::drop` downgrades remaining errors to a logged leak instead of panicking.
- Add deferred-free queue for buffers dropped inside a `with_*` callback; `ContextGuard::drop` flushes it when the borrow ends.
- Centralize stream/pool scheduling in `default_stream_and_pool` and `future_on_default_stream`, replacing 23 copies of the `DeviceFuture`/`ExecutionContext` boilerplate with one map borrow, and attach the device pool at call sites that had missed it (`cutile-macro` `compile()`/`compile_on`, benchmarks, examples).
- Refactor `ExecutionContext::new` to be pure and propagate context errors from pool/cache lookups.

## Test plan

`cargo test -p cuda-async --test pool_allocation` passes  
`./scripts/run_all.sh` passes


elibol:
Looking good! Some feedback / questions:
- Would it work to have the intended contract explicit? Nested `with_global_device_context` access is not supported and should fail fast on the `Borrowed` state.
- `pool_for_stream` and `contains_cuda_function` should return `Result<_, DeviceError>` and propagate context errors.

If I haven't missed something, then let's also add tests so that nested context access fails loudly and pool/cache lookup errors are propagated rather than downgraded. Panic recovery ought to also restores the context state.

goog00:
All three points addressed, split into two commits:

- No-nesting contract — made explicit in the `with_global_device_context*` docs: re-borrowing the per-thread map from a callback panics on Borrowed by design.
- Error propagation — `pool_for_stream` / `contains_cuda_function` now return `Result<_, DeviceError>` and propagate context errors instead of downgrading them to `None` / `false`. This required making `ExecutionContext::new` a pure constructor with a `#[must_use] .with_pool()` builder so the error surfaces at the (already fallible) call sites; updated all ~30 sites + the `cutile-macro `launcher template.
- Regression tests — `nested_pool_lookup_panics_loudly`, `panic_recovery_restores_context_state`, `pool_and_cache_lookups_propagate_poison_error`.

Kept in one PR rather than a follow-up because the error-propagation fix rewrites the same call sites this PR already restructured — splitting would churn them twice and leave main with the contract doc contradicting the code.

PR #147, opened 2026-05-19
feat: process-global single-flight kernel cache + .compile() warmup API

goog00:
## Scope & context
Split out of #81 per review feedback — PR 1 of 3 (process-global in-memory single-flight cache + per-kernel `.compile()` warmup). `execute_warmup` is intentionally not in this PR and lands separately (tracking: #180). Disk-cache persistence is likewise deferred to a follow-up (tracking: #181); per the #81 review it will be opt-in via an explicit API, not default-on with a negative env opt-out. Warmup roadmap: Add Warmup APIs #5

Also addressing #81 review: cache-verification tests assert a test-only hit/miss counter (`jit_compile_count`) instead of wall-clock timing; `run_step` in `run_cpu_tests.sh` includes its label argument.

## Summary
Replaces the per-thread kernel cache with a process-global in-memory cache (shared across all threads in the process; not persisted to disk or across processes), and adds a per-kernel `.compile()` / `.compile_on(stream)` warmup terminal fed by non-allocating `api::meta` tensors, letting callers pre-compile a specialization before its first real launch.

### Unified compile path (`compile_from_context` in tile_kernel.rs)
- The launch path was rewritten from the old thread-local lookup/insert (`contains/get_cuda_function → compile → insert_cuda_function`) to `kernel_cache_slot(key).get_or_try_init(...)` — the same single-flight path warmup uses, so launches and warmup never diverge.
- Takes a `source_hash` parameter (threaded from the macro) and builds the hardened key (`gpu_name / compiler_version / cuda_toolkit_version / device_id`).
- Module ASTs are built **lazily, only on a real cache miss** (not once per batch).
- The old `.expect("Failed to compile function.")` on cubin load is now a proper `Error::KernelLaunch` instead of a panic.

### Global kernel cache (`DashMap<String, Arc<OnceCell<CompiledKernel>>>`)
- `DashMap` gives cross-thread sharing with fine-grained shard locking.
- The inner `OnceCell` (once_cell::sync) implements single-flight dedup: if threads race on the same kernel, only one runs the compile closure; the rest block and receive the same result.
- The shard lock is released before the slow compile begins, so compilation never holds it.
- `AsyncDeviceContext` holds per-thread state (scheduling policy, deallocator stream, mem pool); `Device` handles and the kernel cache are global singletons.
- Caches loaded driver handles (`Arc<Module>`/`Arc<Function>`), not `/tmp` paths — the cubin is `cuModuleLoad`-ed eagerly inside the single-flight closure, so `/tmp` is only transient scratch for the external `tileiras` step.
- **Grows unbounded**: no cap or LRU yet (cutile-python caps at 2 GB with LRU); eviction deferred to the disk-cache follow-up.

### Warmup surface: `api::meta` + `.compile()` / `.compile_on()`
- Warmup is per-kernel and terminal-based (no batch `compile_warmup(specs)`). To warm a specialization, build the exact call you'd launch but pass `api::meta::<T>(&shape)` inputs and end with `.compile()` (or `.compile_on(stream)`) instead of `.sync()`.
- `.compile()` runs the same `execute` path as a launch but stops before the launch — JIT-compiles and caches without allocating or running. It reuses the normal launcher and derives the key from argument metadata, so the warmed entry is byte-identical to what the real `.sync()` looks up.
- `.compile()` warms the calling thread's default-policy device; `.compile_on(stream)` warms the stream's device (mirrors `sync`/`sync_on`), enabling per-device and borrowed interop-stream warmup — `api::meta` takes its device from the `ExecutionContext`.

```rust
let z = api::meta::<f32>(&[1024]).partition([128]);
let x = api::meta::<f32>(&[1024]);
let y = api::meta::<f32>(&[1024]);
kernels::vector_add(z, x, y).generics(["f32", "128"]).compile()?;
```

### Meta tensors: `Storage::{Device, Meta}` split
- `Tensor.storage` is `Arc<Storage>`, either `Device(DeviceBuffer)` (real allocation) or `Meta { len_bytes, device_id }` (no memory). Modeled on PyTorch meta tensors.
- `from_meta` computes the spec against a fixed 16-aligned sentinel (`META_SPEC_PTR`) so a meta tensor's cache key matches a real allocation's.
- `cu_deviceptr()` panics on `Meta` — reached only on a real launch or host/device copy, so `.compile()` (metadata only) accepts meta while `.sync()`/`.await` reject it the moment they read the pointer. `.sync()` on the `api::meta` op itself still succeeds (it only materializes the handle).
- reshape/view/slice/reinterpret recompute the spec through `spec_ptr` (sentinel for meta, real pointer for device) and never dereference, so warming kernels that reshape/slice their inputs neither panics nor diverges from the real key. `DivHint::from_ptr` clamps alignment to 16 and every real allocation is ≥16-aligned, so `sentinel + offset` yields the same `base_ptr_div` as a real `base + offset` **exactly** (both bases are 0 mod 16).

### TileFunctionKey hardening + per-device partitioning
- Added `source_hash / gpu_name / compiler_version / cuda_toolkit_version`: previously the key captured only IR-level parameters, so a toolchain upgrade or source edit could silently reuse a stale in-memory entry.
- Added `device_id`: a loaded `Module`/`Function` handle is bound to the CUDA context of the device it was loaded into, so two identical GPUs (same `gpu_name`) must not share a cache entry — reusing device 0's handle on device 1 is invalid. Restores the per-device partitioning the pre-refactor thread-local cache had implicitly.
- `TileFunctionKeyBuilder` so the key's many fields can't be silently constructed with transposed `String`/vec arguments.
- `_SOURCE_HASH` emitted per module by the macro.

### Observability
- `jit_compile_count()` — process-global `AtomicU64`, +1 per real compile (miss), unchanged on hits. Lets tests assert warmup actually prevented recompilation.
- `CUTILE_JIT_LOG=1` — logs compile start/finish + elapsed to stderr.

## Test Plan
- **Formatting** — `cargo fmt --all -- --check` passes clean across the workspace.
- **warmup.rs (host CPU)** — key correctness: `get_hash_string` distinguishes every field that affects reuse (`source_hash`, `gpu_name`, `compiler_version`, `cuda_toolkit_version`, `device_id`, generics, spec args, compile options).
- **specialization_tests.rs (CPU)** — meta/real spec parity: sentinel matches every real ≥16-aligned ptr; full spec byte-identical for the same layout; sliced meta matches real across offsets.
- **warmup.rs (GPU)** — distinct-key population under concurrency; single-flight dedup fires exactly once under a 4-thread race; meta `view`/`slice` match real specs without panicking.
- **warmup_bench.rs (GPU)** — `.compile()` (meta) warms the cache and the warmed first launch is +0 compile; second-call cache hit; asserts the `jit_compile_count` delta, not wall-clock.

## Results (RTX 4090)
Warmup / cache hits eliminate first-call JIT — measured via `jit_compile_count` (timings shown for context, not asserted):

| Scenario | Cold / first | Cached / warmed | JIT compiles |
| --- | --- | --- | --- |
| In-memory cache (tile=16) | ~283 ms | ~227 µs | 1st +1, 2nd +0 |
| `.compile()` warmup first-call (tile=64) | ~283 ms (no warmup, incl. JIT) | ~259 µs (warmed call) | warmup +1, warmed call +0 |

After warmup the first production call is a sub-millisecond cache hit vs ~280 ms cold JIT, and each kernel compiles exactly once per process.

<details>
<summary>Raw <code>CUTILE_JIT_LOG=1</code> output</summary>

```
╔══════════════════════════════════════════════════════════╗
║       Memory Cache Verification: 1st vs 2nd Call        ║
╠══════════════════════════════════════════════════════════╣
║  First  call (tile=16):    283.4ms  (JIT: +1 compile)  ║
║  Second call (tile=16):    227.3µs  (cache: +0 compile)║
╚══════════════════════════════════════════════════════════╝

╔══════════════════════════════════════════════════════════╗
║         Warmup Verification: First-Call Latency         ║
╠══════════════════════════════════════════════════════════╣
║  Without warmup (tile=32):    282.6ms  (includes JIT)  ║
║  Warmup step     (tile=64):    270.1ms  (pre-compile)   ║
║  With warmup     (tile=64):    258.5µs  (cache hit)     ║
╠══════════════════════════════════════════════════════════╣
║  JIT compiles: cold +1, warmup +1, warmed call +0       ║
╚══════════════════════════════════════════════════════════╝
```
</details>


goog00:
Hi @elibol , just a friendly ping — this PR is ready for review whenever you have bandwidth. Happy to address any feedback or split the changes further if that helps. Thanks!

goog00:
Hi @elibol, friendly ping on this — there are two follow-up commits ([execute_warmup](https://github.com/NVlabs/cutile-rs/commit/a3863f27f4b5a76d2b390be953bc0b22148aa0ea) and [jit disk cache](https://github.com/NVlabs/cutile-rs/commit/f133f4d3ea16a502165b83a514a0979bc162c27b)) waiting on this PR before they can be submitted. Let me know if there's anything I can do to make the review easier. Thanks!

elibol:
Hi @goog00! Thanks for your patients. The last couple of weeks have been a bit  heavy. Reviewing your PRs is a high priority.

goog00:
Hi @elibol, just checking in — any bandwidth to take a look this week? Happy to address any feedback.

elibol:
Thanks @goog00! I'm really excited about this one. I really appreciate your patience and persistence.

Could you link the warmup roadmap issue (#5) and point to the tracking issues for the `execute_warmup` and disk-cache follow-ups? That makes the staging clear from the PR alone.

Do you think the added cost of the cache-hit path is significant? Wondering what it looks like on a small kernel compared to before. Building the hardened key on every launch, the `get_gpu_name()` mutex especially (`cuda_tile_runtime_utils.rs:411`), now happens ahead of the thread-local lookup, and I'm curious if it's measurable. A single-threaded loop that asserts `jit_compile_count` stays flat while timing each launch would show it. If it is, it might make sense to do the thread-local lookup on the cheap key and only build the full key on a miss, and make `get_gpu_name` a `OnceLock`.

The launch and warmup paths duplicate the same compile-and-cache sequence. What do you think about factoring that into one helper that the `compile()` terminal below can reuse too?

The cache has no cap or LRU yet. Can we add at least a note in the code that it grows unbounded (cutile-python caps it at 2 GB with LRU)? Separately, it caches the cubin's `/tmp` path, so a worker thread can fail loading a kernel another thread compiled earlier if `/tmp` was cleaned in between. Low-odds, so not for this PR, but caching the bytes instead of the path makes it moot once the disk cache lands.

On the API, can we replace the warmup surface (`_compile_warmup`, `WarmupSpec`, `EntryMeta`, `_entries`) with a `compile()` terminal on `TileKernel` that compiles and caches without launching, alongside `.sync()`? We'll need to rename the existing internal `compile` function to free the name. Feed it non-allocating tensors, an `api::meta::<f32>(&[1024])` backed by a `Storage::{Device, Meta}` split, modeled on PyTorch's meta tensors, so the existing call works unchanged:

```rust
let z = api::meta::<f32>(&[1024]).partition([128]);
let x = api::meta::<f32>(&[1024]);
let y = api::meta::<f32>(&[1024]);
kernels::vector_add(z, x, y).generics(["f32", "128"]).compile()?;
```

This reuses the builder and cache key, stays typed and per-kernel, and replaces all four warmup symbols. Meta tensors must error on any execution path (`.sync()`, `.await`, `execute_warmup`) and be accepted only by `compile()`, so confirm nothing on the compile path derefs `storage` (the pointer is only read at launch, `tensor.rs:1357`).

For the eventual disk cache, weigh cutile-python's content-addressed key, a hash of the compiled bytecode, which needs no separate source, toolkit, or arch fields and would also remove the per-launch key cost above.

goog00:
> Do you think the added cost of the cache-hit path is significant? Wondering what it looks like on a small kernel compared to before. Building the hardened key on every launch, the `get_gpu_name()` mutex especially (`cuda_tile_runtime_utils.rs:411`), now happens ahead of the thread-local lookup, and I'm curious if it's measurable. A single-threaded loop that asserts `jit_compile_count` stays flat while timing each launch would show it. If it is, it might make sense to do the thread-local lookup on the cheap key and only build the full key on a miss, and make `get_gpu_name` a `OnceLock`.

I measured it. Hardware: RTX 4080 Super, Intel Xeon 8365A (12 vCPUs). Small kernel, tile=128, f32.

I ran the single-threaded cache-hit path (`cache_hit_path_cost`, medians, stable across 6 runs):

```
=== cache-hit-path cost (tile=128, f32) ===
  (A) end-to-end warmed launch (real per-call)           median=173.029µs   100%
  (B) build hardened key + hash (added per-launch CPU)   median=  5.471µs    3.2%
  (C) get_gpu_name() only (mutex + String clone)         median=  0.282µs    0.16%
```

The loop asserts `jit_compile_count` stays flat, so I know every iteration is a genuine cache hit. So I see the cache-hit path adding ~5.5µs/launch (~3% of a ~173µs real call), and the `get_gpu_name()` mutex is only ~0.16% of it — the cost is the key build + hashing, not the lock.

I also checked it under concurrency (`hit_path_contention` — the full hit path across N threads, so the lock sits at its real duty cycle):

```
=== hit-path (build key + hash) contention ===
  threads= 1  per-call=  6.8µs  throughput=  148k/s   (1.0x)
  threads= 2  per-call=  7.0µs  throughput=  282k/s   (1.9x)
  threads= 4  per-call=  6.9µs  throughput=  580k/s   (3.9x)
  threads= 8  per-call=  7.1µs  throughput= 1.12M/s   (7.6x)
  threads=16  per-call= 11.5µs  throughput= 1.28M/s   (8.7x)
```

I see per-call stay flat and throughput scale near-linearly to 8 threads. The dip at 16 is core oversubscription (the box has 12 vCPUs), not the lock — a serializing mutex would have capped throughput at 2–4 threads, not scaled cleanly to 8.

My conclusion: the added cost is measurable but small (~5.5µs, ~3% of a real call), and it's the key build + hash, not `get_gpu_name`. So I'd leave `get_gpu_name` as-is — a `OnceLock` would only save ~0.16%. The other option — doing the cache lookup on a cheap key first and only building the full hardened key on a miss — would remove the whole ~5.5µs, but at ~3% I think it's marginal, and the content-addressed disk-cache key you mentioned later would eliminate it for free anyway, so I'd defer it to that work rather than do it here. I've checked both benches in (`warmup_bench::cache_hit_path_cost`, `::hit_path_contention`) so the numbers are reproducible.

elibol:
@goog00:

1. Run `cargo fmt` across the workspace and commit the result: 7 files currently fail `cargo fmt --check` (device_context.rs, tile_kernel.rs, warmup_bench.rs, warmup.rs, specialization_tests.rs, _module.rs). Also remove the trailing whitespace after `once_cell = "1.19"` in the workspace Cargo.toml.

2. Add a short comment on KERNEL_CACHE in cuda-async/src/device_context.rs stating the cache grows unbounded (no cap or LRU; cutile-python caps at 2 GB with LRU), with eviction deferred to the disk-cache follow-up.

3. Add a `compile_on(stream: &Arc<Stream>)` method next to the generated `compile()`, mirroring the sync/sync_on pair on DeviceOp: identical to compile() but building the ExecutionContext from the given stream instead of `with_default_device_policy(...).next_stream()`. No stream.synchronize() is needed since compile-only enqueues nothing. This enables warmup on borrowed interop streams (see examples/cudarc_interop.rs) and per-device warmup, since api::meta takes its device from the ExecutionContext. Add a doc line on both: compile() warms via the calling thread's default policy; compile_on() warms the stream's device.

4. Fix the api::meta rustdoc in cutile/src/api.rs: `.sync()` on the creation op itself succeeds (it just materializes the meta tensor handle); the panic happens when a kernel launch or host/device copy reads the device pointer. The current wording implies `.sync()` always panics.

5. Rewrite the PR description: it still documents the removed compile_warmup / WarmupSpec / _compile_warmup API and "module ASTs built once per batch". Describe the current surface instead: api::meta tensors + per-kernel .compile() / .compile_on() terminals, Storage::{Device, Meta} split, sentinel-pointer key parity. Also link the warmup roadmap issue (#5) and the tracking issues for the execute_warmup and disk-cache follow-up PRs.

The sentinel-pointer scheme (META_SPEC_PTR = 16), the cache keying, and the panic-on-launch behavior for meta tensors are intentional and were accepted in review. Leave them as-is.

goog00:
Thanks for the thorough review — all addressed.

- **Warmup redesign** — dropped `_compile_warmup` / `WarmupSpec` / `EntryMeta` / `_entries` for a per-kernel `.compile()` terminal fed by non-allocating `api::meta` tensors (`Storage::{Device, Meta}` split); meta tensors panic the moment any execution path reads the device pointer. Internal `compile` renamed. Added `.compile_on(stream)` alongside it, mirroring `sync`/`sync_on`, for interop-stream and per-device warmup. (`24655bd`, `237ecce`, `d446f65`)
- **Shared core** — launch and `.compile()` now go through one `compile_from_context`, so they can't diverge. (`62f3121`)
- **Hit-path cost** — I benchmarked it (`cache_hit_path_cost` `77083b6` / `hit_path_contention` `d98f7ef`): the added per-launch cost is ~5.5µs (~3% of a real call), dominated by the key build + hash, not the `get_gpu_name` mutex (~0.16%) — so I left `get_gpu_name` as-is and deferred the cheap-key-lookup optimization to the content-addressed disk-cache key. 
   The slot lookup itself now takes the read path first (`cache.get`, no alloc), falling back to `entry()` only on a miss (`162ad23`).
- **Unbounded cache** — added the `KERNEL_CACHE` comment (no cap/LRU; cutile-python caps at 2 GB), eviction deferred to the disk cache. We cache loaded `Module`/`Function` handles, not `/tmp` paths, so the path-cleanup concern is moot; bytes land with the disk cache. (`0d08eea`)
- **PR description** — rewritten around the current surface (`api::meta` + `.compile()`/`.compile_on()`, `Storage::{Device, Meta}`, sentinel-pointer key parity).

Follow-ups on the roadmap (#5): `execute_warmup` → #180, opt-in on-disk cubin cache → #181.

Thanks again for the careful review and guidance — happy to keep iterating on any further feedback.

elibol:
Looking good! My main concerns are about what happens when compilation fails. First, compile failures panic through the new `.compile()` API instead of returning an error, and a failure on one thread can race a concurrent retry and orphan its successful compile, so we'd like the failure path cleaned up before merge. Second, the two unused cache-eviction functions can unload GPU code that's still running, so we'd suggest just removing them for now; and third, `.compile()` skips grid validation, so a bad config "warms up" fine and then fails on first real launch. Finally, the cache key has a few gaps (wrong compiler fingerprint, dependency sources not hashed, 64-bit hash as identity) that are harmless in-memory but need resolving before the disk-cache follow-up builds on it. Fine to defer, just please track in #181.

So, before merge:

1. **Make compile failures return `Err` instead of panicking.** `compile_tile_ir_module` (`cutile-compiler/src/cuda_tile_runtime_utils.rs:~364-430`) panics on tileiras non-zero exit, spawn failure, serialization, and bytecode-verification errors. These unwind through `.compile()`'s `Result` and skip the `Err`-arm `remove_if` cleanup in `compile_from_context` (`cutile/src/tile_kernel.rs:~485`), leaving permanent empty slots in the global cache. Suggest converting these to `Result` (like the cubin-load path already was) so the cleanup and the warmup API's error contract both work.
2. **Close the failed-slot eviction race.** In `tile_kernel.rs:488`, `remove_if(&key_str, |_, cell| cell.get().is_none())` can evict a slot while a waiter woken by `once_cell` is retrying the same cell — its successful compile ends up orphaned and single-flight breaks (a later thread starts a duplicate compile). Suggest adding `&& Arc::strong_count(cell) == 1` to the guard; the shard write lock makes the count stable there.
3. **Drop `evict_kernel` / `clear_kernel_cache` for now** (`cuda-async/src/device_context.rs:132-143`). They have no callers, and dropping the cache's `Arc` can `cuModuleUnload` a module whose kernel is still in flight (the launcher releases its `Arc<Function>` at enqueue, not GPU completion — see the TODO at `device_future.rs:126`). Suggest deleting them and reintroducing eviction with the LRU follow-up, designed around a sync or deferred unload.
4. **Run grid validation under `.compile()`.** `infer_launch_grid` sits inside the `if !self._compile_only` gate (`cutile-macro/src/kernel_launcher_generator.rs:~903`), so conflicting partition grids warm "successfully" and fail at the first real `.sync()`. It's pure and cheap — suggest hoisting it above the gate so warmup validates what the launch will.

Before the disk-cache follow-up builds on this key (fine to defer, but please note in #181):

5. **Fingerprint the resolved tileiras, not nvcc.** `get_cuda_toolkit_version` (`cuda_tile_runtime_utils.rs:43-70`) parses `nvcc --version`, but compilation uses a separately resolved tileiras — driver-only machines get `"unknown"`, and `CUTILE_TILEIRAS_PATH` changes don't touch the key. The same file already has accurate machinery (`tileiras_and_toolkit()` + `cuda_version_from_header()` / probe fallback) to reuse.
6. **Note that `_SOURCE_HASH` excludes dependency modules.** It hashes only the module's own source (`cutile-macro/src/_module.rs:348`), but the JIT links dep modules via the use-graph, so a helper-module edit changes the cubin without changing the key. Safe in-memory; would serve stale cubins from disk. Suggest a doc caveat now and dep-closure hashing before persistence.
7. **Consider keying the cache on `TileFunctionKey` itself** instead of its 64-bit `get_hash_string()` (`device_context.rs:127`) — the hash is currently the sole identity, so a collision silently returns the wrong cubin. Pre-existing on main and low probability, but the structured key already derives `Eq + Hash`, so it's a free fix.

Nits:

8. `get_gpu_name` takes a global `Mutex` + `String` clone on every launch, including hits (`cuda_tile_runtime_utils.rs:76-84`) — consider caching per device in a `OnceLock` so the hit path is lock-free.
9. `cutile/Cargo.toml:29` declares `once_cell` but nothing in the crate uses it — remove.
10. `CUTILE_JIT_LOG` only accepts `"1"`, while `setup_diagnostics_enabled` in the same pipeline accepts `true/yes/on` — suggest sharing one parser.
11. `api::meta` casts dims unchecked with `d as i32` (`cutile/src/api.rs:451`); a dim that's a multiple of 2³² silently becomes shape `[0]` and warms the wrong key. A checked conversion would harden this.

goog00:
Thanks for the thorough review — all four before-merge items and the nits are addressed:

- **Compile failures return `Err`** — `compile_tile_ir_module` now returns `Result`; the tileiras exit, spawn, serialization, and verification paths propagate through `.compile()`.
- **Eviction race** — guard is now `cell.get().is_none() && Arc::strong_count(cell) == 1` (drop our own `slot` first so the count reflects only the map).
- **`evict_kernel` / `clear_kernel_cache`** — removed; will reintroduce with the LRU follow-up around a safe unload.
- **Grid validation** — `infer_launch_grid` hoisted above the `!_compile_only` gate, so warmup validates the grid too.

Nits: `get_gpu_name` now caches per device in a `OnceLock` (lock-free hits); dropped the unused `once_cell` dep; `CUTILE_JIT_LOG` shares one parser with the diagnostics flag; `api::meta` uses a checked `i32` conversion.

The three cache-key gaps are tracked in #181 for the disk-cache follow-up.

Thanks again for the careful review and guidance — happy to keep iterating on any further feedback.

elibol:
Re-reviewed the two fix commits — all four before-merge items are properly addressed (verified the `JITError` propagation chain, the `drop(slot)` + `strong_count == 1` eviction guard, the evictor removal, and the hoisted grid validation), and the nits look good too. Thanks for the quick turnaround!

One tiny thing: `Cargo.lock` wasn't regenerated after dropping the `once_cell` dep — I've pushed that one-line lockfile update directly to your branch (787bf78) to save a round trip. Hope that's okay!

LGTM once CI is green.

PR #149, opened 2026-05-22
fix tutorial imports and resolve type mismatches in examples

vito-lin-dev:
Hi,

This PR is to fix several compilation issues and misleading import paths in the cutile tutorials and examples.

**changes:**

**1. Rust Dependencies Hint**: Corrected the dependencies  to let learner know to install candle-core by `cargo add candle-core`

**2. Type Mismatch Fix**: Resolved a type mismatch error in the Softmax example where `i32` was used for partition() dimensions instead of `usize`.

**3. Refactor/Cleanup**: Cleaned up dependency handling and type usage across the examples to ensure consistent compilation with current Rust standards.

I have verified these changes within the cutile-examples environment, and they resolve the build failures I encountered.

Thanks for maintaining this project!

### DIFF
--- a/cutile-book/tutorials/04-matrix-multiplication.md
+++ b/cutile-book/tutorials/04-matrix-multiplication.md
@@ -57,13 +57,14 @@ cargo add candle-core
 ```
 
 ```rust
-use cuda_async::device_operation::DeviceOp;
-use cuda_core::Device;
+use cutile::cuda_async::device_operation::DeviceOp;
+use cutile::cuda_core::Device;
 use std::sync::Arc;
 use cutile;
 use cutile::api;
-use cutile::candle_core::WithDType;
+use candle_core::WithDType;
 use cutile::error::Error;
+use cutile::prelude::PartitionOp;
 use cutile::tensor::{IntoPartition, Tensor, ToHostVec, Unpartition};
 use cutile::tile_kernel::TileKernel;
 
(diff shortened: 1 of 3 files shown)


vito-lin-dev:
Dear maintainer, I'd like to provide an update on this PR.

I have verified that these changes correctly resolve the library path issues by ensuring the project compiles successfully and generates valid .cubin files (as seen in the compile_only examples). 

**Please note that I do not have a physical GPU available in my environment**. Consequently, I am unable to pass the full `bash scripts/run_all.sh test suite, as it triggers DriverError 

```
DriverError(303, "CUDA driver library unavailable: CUDA driver too old: built against 13.2 but runtime is 12.2"), 
```

```
failed to get CUDA device: DriverError(100, "no CUDA-capable device is detected")).
```

I have ensured that all CPU-side unit tests and compilation-only regression tests pass locally. I would appreciate it if the maintainers or someone else could `verify the GPU runtime tests via the project's CI/CD pipeline to ensure that no regressions were introduced` or `suggest me what to do`. Thank you for your time and guidance!

elibol:
@vito-lin-dev thanks for these, and sorry they sat so long. The tutorials and examples have been rewritten since May (they are now compiled and executed as tests, and the import layout changed with #248), so these fixes no longer apply to main. I am going to close them rather than ask you to redo the work; if anything in the current book is broken, an issue with the snippet is the fastest way to get it fixed.

Issue #148, opened 2026-05-22
Decouple cutile from cuda-core to allow external crate integration


vito-lin-dev:
**Subject:** 
Request for better modularity between **cutile** and **cuda-oxide/cuda-core**

**Description:**
I am currently exploring cutile and practicing the examples. I noticed that the tutorials rely on cutile being part of a specific workspace environment (using workspace = true in Cargo.toml).

While this works great within the cutile-rs repository, it makes it quite difficult to use cutile in an external project. When I try to use cuda-core or cuda-oxide (the underlying driver crates) alongside cutile, I encounter type mismatch errors (e.g., expected &Arc<cuda_core::Stream>, found &Arc<cutile::cuda_core::Stream>).

**The Problem:**
Currently, cutile seems to strictly require its own internal wrapper types of CudaContext and Stream. This forces developers to work exclusively within the cutile-rs workspace, as using external cuda-core instances leads to type incompatibility.

**My Questions:**
1. Are there any plans to make cutile more compatible with external cuda-core (or cuda-oxide) instances?
2. Tutorial/Documentation: Are there any recommended patterns for using cutile as a library in an external project without cloning the entire workspace?

vito-lin-dev:
Hi,

I was following the [Hello World tutorial](https://www.google.com/url?sa=E&q=https%3A%2F%2Fnvlabs.github.io%2Fcutile-rs%2Ftutorials%2F01-hello-world.html) and encountered some compilation errors due to type mismatches between `cuda-core` and `cutile`.

It appears that the import paths suggested in the tutorial are misleading. The tutorial currently recommends:

```Rust
use cuda_async::device_operation::DeviceOp;
use cuda_core::Device;
```

However, this causes type mismatch issues because there I don't find the message tell me not to use the external cuda-core crate instead of the one re-exported by cutile. The code compiles correctly if I use the paths provided by the cutile crate:

```Rust
use cutile::cuda_async::device_operation::DeviceOp;
use cutile::cuda_core::Device;
```

I have verified this in the cutile-examples environment, and using these specific imports allows the code to run as expected.

Updating the tutorial to reflect these paths would be very helpful for new users to avoid this confusion. 

I want to submit a PR to update the whole documentation/tutorial and example. 

**Please let me know if the project benefit from this change, thanks!**

Thanks for the great work on this project!

vito-lin-dev:
Dear maintainer, I'd like to provide an update on this PR.

[...]

PR #150, opened 2026-05-22
fix: change the improper import of cutile/cuda_async and cutile/cuda

vito-lin-dev:
## Description

This PR addresses the issue reported in:
https://github.com/NVlabs/cutile-rs/issues/148

## Problem
The tutorial and example code currently use imports such as:
- `use cuda_async::*`
- `use cuda_core::*`

This can be misleading for new users, as it is not clear that these dependencies are intended to be used within the Cutile environment. As a result, users may incorrectly configure their `Cargo.toml` (e.g., missing `workspace = true`) or attempt to pull dependencies directly, leading to build errors.

## Changes
This PR updates the import paths to use the Cutile re-exported modules:

- `use cutile::cuda_async::*`
- `use cutile::cuda_core::*`

## Impact
- Clarifies the intended usage of the Cutile environment
- Reduces confusion for new users following tutorials
- Prevents incorrect dependency setup in `Cargo.toml`

### DIFF
--- a/cutile-book/tutorials/01-hello-world.md
+++ b/cutile-book/tutorials/01-hello-world.md
@@ -11,8 +11,8 @@ Tile kernels are functions which run as `N` copies concurrently and in parallel 
 Here is a kernel that prints "hello" from the GPU:
 
 ```rust
-use cuda_async::device_operation::DeviceOp;
-use cuda_core::Device;
+use cutile::cuda_async::device_operation::DeviceOp;
+use cutile::cuda_core::Device;
 use cutile;
 use cutile::error::Error;
 use cutile::tile_kernel::TileKernel;
--- a/cutile-book/tutorials/02-vector-addition.md
+++ b/cutile-book/tutorials/02-vector-addition.md
@@ -7,8 +7,8 @@ In cutile, tile threads run concurrently and each tile knows its coordinates via
 ---
 
 ```rust
-use cuda_async::device_operation::DeviceOp;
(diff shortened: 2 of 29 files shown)


vito-lin-dev:
Dear maintainer, I'd like to provide an update on this PR.

I have verified that these changes correctly resolve the library path issues by ensuring the project compiles successfully and generates valid .cubin files (as seen in the compile_only examples). 

**Please note that I do not have a physical GPU available in my environment**. Consequently, I am unable to pass the full `bash scripts/run_all.sh test suite, as it triggers DriverError 

```
DriverError(303, "CUDA driver library unavailable: CUDA driver too old: built against 13.2 but runtime is 12.2"), 
```

```
failed to get CUDA device: DriverError(100, "no CUDA-capable device is detected")).
```

I have ensured that all CPU-side unit tests and compilation-only regression tests pass locally. I would appreciate it if the maintainers or someone else could `verify the GPU runtime tests via the project's CI/CD pipeline to ensure that no regressions were introduced` or `suggest me what to do`. Thank you for your time and guidance!

elibol:
@vito-lin-dev thanks for these, and sorry they sat so long. The tutorials and examples have been rewritten since May (they are now compiled and executed as tests, and the import layout changed with #248), so these fixes no longer apply to main. I am going to close them rather than ask you to redo the work; if anything in the current book is broken, an issue with the snippet is the fastest way to get it fixed.

PR #169, opened 2026-06-19
Fix device op Send bounds

fallintoplace:
## Summary

- Remove the unconditional `Send` impl for `Value<T>` so it follows Rust auto-trait rules.
- Require `Empty` closures to be `Send` and remove its manual `Send` impl as well.

## Root Cause

`Value<T>` wrapped an arbitrary payload but declared `unsafe impl<T> Send`, allowing non-`Send` payloads such as `Rc<_>` to move across threads. While the `DeviceOp` impl was bounded by `T: Send`, the public wrapper type itself had a broader auto-trait contract than its field.

The nearby `Empty<O, DO, F>` wrapper had the same issue for its closure field: it manually implemented `Send` without requiring `F: Send`.

## Unsafe Send/Sync Audit

- Fixed: `Value<T>` and `Empty<O, DO, F>` had over-broad `Send` impls.
- Looks intentionally bounded: `AndThen`, `Zip`, `SelectLeft`, `SelectRight`, `AndThenWithContext`, `DeviceOperationPartition`, `UnwrapPartition`, and `TensorToHostVec` all require `DeviceOp` or `Send` bounds on their payloads.
- Opaque CUDA handle wrappers: `Device`, `MemPool`, `Stream`, `Module`, `Function`, `DeviceBuffer`, `DevicePointer`, `AsyncKernelLaunch`, and `GraphLaunch` are manual handle/pointer wrappers; they still deserve invariant docs, but I did not find the same unbounded generic-payload bug there.
- Needs follow-up documentation/review: `SharedExec` and `Select*` use `UnsafeCell` with custom execution invariants. Their bounds are not the same bug, but they should have explicit safety comments.

## Validation

- `cargo fmt --check`
- `git diff --check`
- `cargo check -p cuda-async` blocked locally because `CUDA_TOOLKIT_PATH` is unset and no CUDA 13.2+ toolkit exists at the default `/usr/local/cuda*` paths.


elibol:
Nice catch and thanks for the contribution! I added a few tests to avoid regression. Otherwise LGTM!

PR #179, opened 2026-07-02
fix: block on in-flight stream work when dropping a pending DeviceFuture

elibol:
## Summary

Dropping a `DeviceFuture` after its first poll freed the pending result while the stream work producing it could still be in flight: a `Tensor`'s device buffer would be freed — and become recyclable by the pool — while a kernel still writes it. Future cancellation is ordinary safe code with async runtimes (`tokio::select\!`, dropped `JoinHandle`s, timeouts), so this was memory unsafety reachable from safe code.

- Add a `Drop` impl that synchronizes the execution stream when the future is dropped in the `Executing` state, and document the cancellation semantics (dropping a pending future blocks until the underlying stream work completes).
- Null-check host allocations before `Vec::from_raw_parts` on the device-to-host copy paths (previously passed a null pointer on allocation failure).
- Regression tests: a deterministic gate test (blocks the stream with a host function, polls once, drops on another thread, asserts the drop cannot return while the stream is busy — verified to fail without the fix), a drop-pending stress test, and an idle-drop unit test.

Full GPU test suite passes (27/27, RTX 5090).

### DIFF
--- a/cuda-async/src/device_future.rs
+++ b/cuda-async/src/device_future.rs
@@ -7,12 +7,13 @@
 
 use crate::device_operation::{DeviceOp, ExecutionContext};
 use crate::error::DeviceError;
-use futures::task::AtomicWaker;
+use futures::task::{waker, ArcWake, AtomicWaker};
 use std::future::Future;
 use std::pin::Pin;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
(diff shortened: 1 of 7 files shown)


elibol:
Superseded by #258 (447e906), which landed the root `DeviceFuture` `Drop` that this PR proposed a blocking variant of. Closing.

PR #183, opened 2026-07-06
fix flake's tileiras failing and add aarch64 support

Firestar99:
close https://github.com/NVlabs/cutile-rs/issues/172

NOTE: This change will make macos targets built for aarch64 targets by default instead of x86 ones.

Issue #172, opened 2026-06-23
nix flake broken


Firestar99:
### Operating System

"NVIDIA DGX OS" aka. customized Ubuntu 24.04 aarch64

### GPU and GPU Architecture

sm_121 (DGX Spark Blackwell)

### Error Description

I'm trying to run any example of cutile on an NVidia [DGX Spark](https://www.nvidia.com/en-us/products/workstations/dgx-spark/), which is an aarch64 CPU with a blackwell iGPU. 

First let me rant a bit on how outdated the software of your ubuntu-based "official DGX OS" is. It first fails to build due to not finding `stddef.h`. Luckily, cuda-oxide [has a section on that in their readme](https://github.com/NVlabs/cuda-oxide#clang-host-cuda-bindings) and tells you to `apt install clang-21`...
```
$ sudo apt install clang-2
clang-20           clang-20-doc       clang-20-examples
```
These boxes don't have anything more up-to-date than llvm-20... (I'm getting more and more tempted to install nixos on that thing)

Anyways, you have a nix flake so I tried that, which compiles but then fails with:
```
Error: error: DriverError(303, "CUDA driver library unavailable: failed to load any of [\"libcuda.so.1\", \"libcuda.so\"]: libcuda.so: cannot open shared object file: No such file or directory")
```
Looking at your nix flake, it was written only for x86_64, so [I patched it to work for aarch64](https://github.com/NVlabs/cutile-rs/compare/main...Firestar99:cutile-rs:aarch64) (I can PR that later, but note that Mac is also aarch64 so it'll now build binaries for aarch64 by default). 

Running `cargo run --example saxpy` within that devshell I get:
```
thread 'main' panicked at cutile-compiler/src/cuda_tile_runtime_utils.rs:185:9:
/nix/store/6b37zanr0lsl0h3pln1rm6gmsnchss4l-cuda-toolkit-13.3/bin/tileiras failed while compiling Tile IR bytecode.
status: exit status: 5
command: /nix/store/6b37zanr0lsl0h3pln1rm6gmsnchss4l-cuda-toolkit-13.3/bin/tileiras --gpu-name sm_121 --opt-level 3 -o /tmp/nix-shell.LHqlxm/621a58bf-5fe3-44a5-8dd4-b2f39af61c1d.cubin /tmp/nix-shell.LHqlxm/621a58bf-5fe3-44a5-8dd4-b2f39af61c1d.bc
target gpu: sm_121
[...]

PR #185, opened 2026-07-06
fix: harden nix cuda-toolkit assembly and derive driver dirs from host arch

elibol:
## Summary

Follow-ups to #183 (issue #172):

- **Guarded archive merging**: the six redist archives now extract with `--skip-old-files --warning=existing-file`, so cross-archive path collisions are deterministic (first-wins) and visible in the build log instead of silently taking whichever archive extracted last. `libnvvm` moves ahead of `cuda_nvcc` in the list (both can populate `nvvm/`), and the order-matters constraint is documented.
- **Assembly assertions**: the toolkit build now fails loudly if `bin/tileiras` or `nvvm/lib64/libnvvm.so` is missing — the exe-relative-lookup invariant from #172 can no longer silently regress on a version bump.
- **Host-arch driver dirs**: the driver-library search derives the Debian multiarch directory from `uname -m` instead of enumerating per-arch triplets — shorter, searches the host arch first, and covers future architectures without edits.
- **Hash bookkeeping**: comment points at NVIDIA's `redistrib_13.3.0.json` manifest (all 12 pinned hashes were verified against it during the #183 review).

Verification note: no nix installation was available on the dev machine, so these changes are review-verified only — a `nix develop -c cargo run -p cutile-examples --example hello_world` smoke on an x86_64 box (and ideally @Firestar99's Spark) before merge would be appreciated.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -37,6 +37,14 @@
             url = "${cudaRedistBase}/${name}/${cudaRedistArch}/${name}-${cudaRedistArch}-${version}-archive.tar.xz";
             sha256 = sha256.${cudaRedistArch};
           };
+        # Hashes come from NVIDIA's manifest for this release:
+        # https://developer.download.nvidia.com/compute/cuda/redist/redistrib_13.3.0.json
+        # (per package: .<name>.<arch>.sha256). Update from there on version bumps.
+        #
+        # ORDER MATTERS: archives are extracted with --skip-old-files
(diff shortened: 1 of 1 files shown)


elibol:
@Firestar99 Would you mind trying out this branch? I'm planning to merge it later this week.

Firestar99 (approved):
Works on my spark!

PR #188, opened 2026-07-08
Use overlap analysis on eq/ne in `bounds_from_bop`

ivarflakstad:
Noticed what I believe is a bug while reading through the code trying to understand the intricacies.

For `eq` and `ne`
```rust
start = min(op(a.start, b.start), op(a.start, b.end), op(a.end, b.start), op(a.end, b.end))
end = max(op(a.start, b.start), op(a.start, b.end), op(a.end, b.start), op(a.end, b.end))
```
can become 
```rust
start = 0
end = 0
```
which could then be interpreted as dead code by the compiler (if I'm not mistaken).

`lt`, `gte`, etc do not have this issue.

### DIFF
--- a/cutile-compiler/src/bounds.rs
+++ b/cutile-compiler/src/bounds.rs
@@ -319,8 +319,28 @@ pub fn bounds_from_bop(op: &TileBinaryOp, a: &Bounds<i64>, b: &Bounds<i64>) -> O
             TileBinaryOp::Add => *a + *b,
             TileBinaryOp::Sub => *a - *b,
             TileBinaryOp::Mul => *a * *b,
-            TileBinaryOp::Eq => bop_bounds(a, b, |a, b| (a == b) as i64),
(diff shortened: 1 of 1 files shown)


elibol:
Thanks, @ivarflakstad! Nice catch.

Do you mind adding a regression test? Cases for disjoint mapping to 0, overlap mapping to [0,1], and exact/exact would be a good start.

I think remainder and bitwise ops  (&, |, ^) have the same issue this PR is fixing. Do you mind fixing those as well? 

For bitwise, for a, b in [0,2], 2 ^ 1 = 3, but the four corners only produce {0, 2}, so corner min/max yields [0,2] and drops the reachable 3. For remainder, [0,3] % 3 gives corners [0,0] but the true range is [0,2]. It also lacks a divide-by-zero guard (unlike div), so a % b with 0 in b's range panics. You can leave the division-by-zero issue to me if you prefer to keep this PR focused on the range analysis.

PR #190, opened 2026-07-12
Add /opt/cuda to default candidates

svenstaro:
Arch Linux uses /opt/cuda and I think it would be neat if it just worked out of the box.


elibol:
Hi @svenstaro! Thanks for the contribution and your patience. Happy to support Arch out of the box. Do you mind having a look a the following?

1. The same candidate list is duplicated in `cutile-compiler/src/cuda_tile_runtime_utils.rs` (`default_cuda_toolkit_candidates`), where it drives runtime discovery of the `tileiras` binary. Without adding `/opt/cuda` there too, an Arch install would pass the build but still fail to locate `tileiras` at run time. Can you mirror the addition there (in the same position, after `/usr/local/cuda`)?
2. The README's "Configure Environment" section lists the default search locations — could you add `/opt/cuda` to that sentence as well?

svenstaro:
Hey @elibol, sorry, didn't see your comment. I think I addressed your comments now.

Issue #181, opened 2026-07-02
Add persistent on-disk kernel cache


goog00:
Part of the warmup roadmap #5. Follow-up to #147.

#5 also calls for persisting the JIT to disk, with the storage abstraction as a trait behind an object-store-like interface (get/put by key). The in-memory cache from #147 is per-process, so every fresh process re-runs the JIT from scratch.

This tracks that on-disk cache: a cold process can reuse a previously compiled kernel instead of recompiling, behind the object-store trait so the backing store is swappable. It should be off by default and enabled only through an explicit API — not on-by-default with a negative env opt-out (raised on #81) — and, per the #147 discussion, safe to share across processes and bounded in size (cutile-python caps its cache at 2 GB with LRU).

Per the #147 discussion, weigh a content-addressed key here — a hash of the compiled bytecode, as cutile-python does — instead of the input-derived key the in-memory cache assembles: it needs no separate source, toolkit, or arch fields and would also drop the per-launch cost of building that key. Storing the compiled bytes rather than a temp path is what lets it work safely across processes.

The #147 review flagged three cache-key issues that fit better here: all are fine while the cache is in-memory and per-process, but each turns into a correctness bug once we persist to disk.

1. **Fingerprint the resolved tileiras, not nvcc.** `get_cuda_toolkit_version` (`cuda_tile_runtime_utils.rs:43-70`) parses `nvcc --version`, but compilation uses a separately resolved tileiras — driver-only machines get `"unknown"`, and `CUTILE_TILEIRAS_PATH` changes don't touch the key. The same file already has accurate machinery (`tileiras_and_toolkit()` + `cuda_version_from_header()` / probe fallback) to reuse.
2. **Note that `_SOURCE_HASH` excludes dependency modules.** It hashes only the module's own source (`cutile-macro/src/_module.rs:348`), but the JIT links dep modules via the use-graph, so a helper-module edit changes the cubin without changing the key. Safe in-memory; would serve stale cubins from disk. Suggest a doc caveat now and dep-closure hashing before persistence.
3. **Consider keying the cache on `TileFunctionKey` itself** instead of its 64-bit `get_hash_string()` (`device_context.rs:127`) — the hash is currently the sole identity, so a collision silently returns the wrong cubin. Pre-existing on main and low probability, but the structured key already derives `Eq + Hash`, so it's a free fix.

PR #193, opened 2026-07-14
feat(jit): opt-in persistent on-disk cubin cache

goog00:
Implements #181: an opt-in, persistent on-disk cache for compiled cubins, so a fresh process skips the `tileiras` backend for kernels an earlier process already compiled. Off by default; enabled with `cutile::jit_cache::enable_default()` or `enable(store)`.

The eviction backend was the one open choice; per review it's the plain-file store below, not SQLite. Ready for review.

## Where the cache sits

The key is a SHA-256 over the serialized bytecode plus target, opt level, and the `tileiras` fingerprint. That puts the lookup inside stage 2 — after the frontend has produced the `.bc`, before `tileiras` is spawned — rather than in front of the whole pipeline keyed on the launch arguments.

This is what #181 asks for ("content-addressed key — a hash of the compiled bytecode"), and it buys correctness by construction rather than by enumeration: the `.bc` already inlines every dependency module, so a changed helper changes the key without a dependency-closure hash, and a changed frontend changes the key without a compiler-version field. (The `_SOURCE_HASH` attribute is consequently not part of the key; its known gap — it doesn't cover dependency modules — is documented for whoever needs it later, but doesn't affect the cache.)

The load-bearing assumption is that bytecode serialization is deterministic: the same module must serialize to identical bytes across processes, or a writer stores under key A while a reader computes key B and every cross-process lookup misses (safe, but useless). The writer holds this by construction — only insertion-ordered maps feed the output — and `cutile-ir/tests/bytecode_determinism.rs` pins it: the test fails the moment any `std::collections::HashMap` is iterated into the bytecode.

The cost is a lower ceiling — a hit still pays the frontend. Measured on two machines (sm_89 and sm_120, release builds, 18 kernel samples from `vector_add` up to `gemm<f16,256,256,64>`), `tileiras` is 53.6%–99.5% of a cold compile, so a hit skips the large majority of it: the `gemm` goes from 16.8 s to 98 ms on sm_89 and 4.0 s to 54 ms on sm_120. The residual frontend cost is 37–56 ms per kernel. Per-sample breakdown available on request.

## Against the issue's requirements

- **Trait, object-store shaped** — `JitStore` with `get`/`put`/`contains`/`delete`/`clear`, keyed by 64-char hex. contains has a get-based default so custom backends stay minimal.
- **Off by default, explicit API only** — no env var can turn it on.
- **Content-addressed key** — above.
- **Size-managed** — 2 GiB default soft target with LRU eviction; transient overshoot is possible. See the eviction backend section below.
- **Stores bytes, not paths** — `compile_tile_ir_module` now returns `Vec<u8>`; this also fixes the leaked `/tmp/<uuid>.cubin` per JIT, and the pre-existing hazard of caching a `/tmp` path that another thread later fails to load.
- **`tileiras` fingerprint, not `nvcc`** — the fingerprint is `tileiras --version` from the same binary the compile resolves (~5 ms, once per process); the old `nvcc --version` parse could report `"unknown"` on a driver-only box and never changed when `CUTILE_TILEIRAS_PATH` did.
- **No 64-bit hash as identity** — the in-process cache is now keyed on `TileFunctionKey` itself. The kernel-cache APIs move from `device_context` to `tile_kernel`, and `get_hash_string()` is renamed to the display-only `display_hash()`.
- **Custom backends** — examples/jit_custom_store.rs implements a non-filesystem JitStore, with get/put/delete annotated against their S3 and SQL equivalents.

## Validation, and what it does not cover

Entries carry a header with `sha256(payload)`, `sha256(bc)`, gpu name, opt level and fingerprint, all re-checked on read.  Torn, corrupted, malformed, or request-mismatched entries are treated as misses and deleted, so they cost one recompile rather than serving an invalid result. A cubin that passes the header check but is still rejected by `cuModuleLoadData` is evicted and recompiled once before the error propagates.

These checksums are integrity, not authenticity: anyone who can write the cache directory can write a valid header. The trust boundary is directory permissions. `enable_default()` creates its root (`$XDG_CACHE_HOME/cutile/kernels`, falling back to `~/.cache/cutile/kernels`) with mode `0o700` and re-tightens the mode if the directory already existed looser, so another user cannot plant entries. A custom root passed to `enable(store)` is trusted as-is — pointing it at a shared or world-writable directory extends trust to everyone who can write there.

## Eviction backend (decided: plain files)

The issue left the backing store open. Decided in review: plain files, with the trait keeping SQLite/S3/etc. available as out-of-tree JitStore impls. The comparison below is the rationale.

**SQLite** (what cutile-python does): one row per entry, cubin as a blob; a hit UPDATEs the last-used time, eviction is `DELETE ... ORDER BY atime`. Exact LRU, correct size accounting, indexed eviction that only touches the rows it removes, concurrency handled by the engine. The costs: SQLite locks the whole database for any write and every hit *is* a write (the atime UPDATE), so hits across processes serialize; a corrupt database file loses the entire cache; and in Rust this means adding rusqlite + bundled C, where Python gets `sqlite3` from the stdlib for free.

**Plain files** (what this PR implements): entries at `<root>/<ab>/<key>.cubin`, sharded by key prefix. A hit touches the file's mtime; eviction scans the directory under a lock file and deletes oldest-mtime first. No shared lock on the hit path, a broken entry only loses itself, no new dependencies (std `File::try_lock`, MSRV 1.89). The costs: LRU is approximate — mtime is a stand-in for "last used" that outside tools can rewrite — and eviction needs a full directory scan, though it's amortized to one scan per `capacity/16` bytes written and only ever runs on the miss path. ccache does exactly this: sharded files, size cap, oldest-timestamp-first.

Correctness is identical either way, per the entry validation above, so a bad eviction costs one recompile in both. One thing that shifted the weights: if the cache can end up on a network filesystem — cluster `$HOME` is often NFS-mounted, and one shared cache for a multi-rank launch is something people will want — then both get worse. SQLite's docs warn its locking is unreliable there; the directory scan becomes a round-trip storm.

Weighed as: a bundled-C dependency paid by every build, versus approximately-right eviction whose worst case is one recompile — resolved toward the latter. See examples/jit_custom_store.rs for the SQL/S3 extension path.

## Review order

1. The `JitStore` trait and the on-disk entry format.
2. Key derivation and the `tileiras` fingerprint.
3. `FileSystemJitStore` and the mtime LRU — the swappable default behind the trait.


elibol:
Let's keep the storage backend as a trait and provide disk-based store as the default.

I'm a bit less optimistic about taking on rusqlite. SQL, Amazon S3, and similar backends could instead be covered by an example of implementing JitStore (an examples/ file or doc section) rather than as dependencies. This seems to be the solution this problem looks like wherever it's solved well: sccache (storage trait, disk default, cloud backends optional), ccache, cargo's caches; on the Python side, joblib's StoreBackendBase with a filesystem default and S3 as a documented extension, and pip's BaseCache/FileCache. It also matches the std idiom: An io::Read/Write-style minimal byte-oriented trait, one well-understood default backend, everything else living outside.

My read on the trade-off you posed: Approximate LRU is the cheap cost, since your validation layer already bounds a bad eviction at one recompile, while a bundled C dependency is paid by every build forever. NFS seems to cut against SQLite harder than against files, per SQLite's own locking docs.

One thing I'd fix before this leaves draft: The GC trigger draws its randomness from the first 8 bytes of a v4 UUID, but byte 6's version nibble is fixed at 4, which lands at bits 52–55 of the LE u64. Draw >= 2^54 always, and at the default capacity any entry <= 128 KiB can never trigger a scan. A workload of small kernels would grow the cache unbounded, which is exactly what the amortization argument promises can't happen.

One more naming thought: "GC" reads as memory-management (tracing, reachability) rather than what this is (cache eviction). Consider `evict/eviction_watermarks/.eviction.lock` over `collect_garbage/gc_watermarks/.gc.lock`.

Smaller suggestions, non-blocking:
- Consider a default impl for `contains` via `get`. The engine never calls it, and it would keep custom backends minimal.
- `DirBuilder::mode(0o700)` could close the brief window between `create_dir_all` and `set_permissions` in `default_location`.
- Maybe name the module `jit_cache` to match the public re-export.

Otherwise LGTM.

goog00:
Thanks , very helpful.

Backend decision adopted: `JitStore` trait with the filesystem store as default,no rusqlite. Added `examples/jit_custom_store.rs` (get/put/delete annotated against their S3 and SQL equivalents) so cloud/SQL backends are an implementation,not a dependency. Agreed on the trade-off read too.

**UUID bug — confirmed and fixed.**  Fix extracts `eviction_draw()`, XOR-folding the UUID's two 64-bit halves so each fixed bit lands on an independent random bit; the 64 result bits are then uniform.
`eviction_draw_has_no_fixed_bits` asserts it directly with real UUIDs at the real draw width, so a future "simplify back to `bytes[..8]`" fails loudly.

**Naming** — renamed throughout: `collect_garbage → evict`, `gc_watermarks → eviction_watermarks`, `.gc.lock → .eviction.lock`, and the module `jit_store → jit_cache`.

**Non-blocking, all done:**
- `contains` now defaults via `get`; `FileSystemJitStore` keeps its `stat` override.
- `default_location` creates the dir `0700` via `DirBuilder::mode`.

Verified on sm_120: `cargo build --workspace --examples`,`cargo test -p cutile-compiler`, both examples run, `jit_disk_cache` twice shows the second process at `stage2_source=disk` with zero backend compiles.

goog00:
One design point I'd like your call on before this merges: cross-process compilation dedup.

Within a process, `OnceCell` ensures a kernel is compiled once. Across processes there is no equivalent: N processes cold-starting together (torchrun/MPI-style, one per GPU) all miss the same key, each spawns tileiras, each writes. This is correct — writes are atomic tmp+rename and content-addressed, so the store converges to one identical entry — but the very first launch wastes N−1 compiles (~16 s each for the largest gemm). Every subsequent start hits.

The obvious fix is a per-key blocking file lock around the miss path, and I deliberately left it out: a slow-but-alive lock holder stalls all waiters with no way to tell "compiling" from "hung", and `flock` is unreliable on NFS — which is where a shared cache is most likely to live. Callers can also sidestep it entirely (one warmup process before the parallel launch).

Options as I see them: **(a)** ship as-is and document the limitation, revisit if first-launch stampede shows up as a real cost; **(b)** add the per-key lock in this PR; **(c)** as-is now, tracking issue for the lock. I lean **(a)/(c)**, but it's a maintainer call on which cost the project would rather carry. Happy to implement **(b)** here if you'd rather have it.

elibol:
Sorry for the delay. Option (c): merge as-is, tracking issue for the stampede.

I agree re: leaving the per-key lock out. A slow-or-dead lock holder stalls every waiter, and flock is weakest exactly where a shared cache lives (NFS). The stampede is one cold start, converges by construction, and has an easy workaround (one warmup process before the parallel launch). Please add a sentence about that to the "Persistent Disk Cache" section this PR adds to guide/jit-compilation.md. 

The tileiras resolution caching LGTM.

I ran the GPU suite locally: All four disk-cache tests pass, including cross-process.

One nit: the cross-process test picks its children by a hardcoded test-name filter, so a rename would make them run zero tests and pass vacuously. Asserting the writer's dir is non-empty would close that.

goog00:
Thanks — I went with option (c) and opened #205 to track cross-process compilation deduplication.

I added the cold-start stampede limitation and single-process warmup workaround to the “Persistent Disk Cache” section in `cutile-book/guide/jit-compilation.md`.

I also hardened the cross-process test by asserting that the writer populates its cache directory. I reproduced the stale-filter case by changing the hardcoded child test name; it now fails explicitly instead of passing with zero tests.

The `tileiras` resolution caching is unchanged. `cargo fmt --all -- --check` passes, and thanks for verifying the full GPU suite locally.

elibol:
Thanks for #205 and the doc/test hardening. All that looks good!

A final pass turned up two more things I'd like to fix before merge, plus small items:

1. The four `jit_disk_cache` tests assert on process-global state (the installed store, the compile counters, one clears the kernel cache), but they're registered in the shared parallel gpu binary. `cache_test_lock` only serializes them against each other, so any concurrent test's first launch can land in the measured window and flake the counter/file-count assertions. That's the race class `warmup_suite.rs` exists for (see its header). Please move the mod `jit_disk_cache` registration there, and also add `cargo test -p cutile --test warmup_suite` to `scripts/run_gpu_tests.sh`. The script doesn't run that suite today, and it's what I use to run GPU tests.

2. `default_location()` tightens a pre-existing loose cache root to 0700 but keeps its contents. Entries planted before the tightening still validate (the header checksum is integrity, not authenticity) so a pre-existing group/other-writable root should be rejected with an error telling the user to delete it, rather than chmodded and reused. cutile always creates the dir 0700 itself, so this only fires on roots some other actor created, but a hit executes device code, so let's close it.

3. Book page: "a hand-planted file … reads as a miss" is only true of garbage. A crafted entry with a recomputed checksum passes validation. Please align that sentence with the module docs (integrity vs. authenticity) and add a line that custom/shared store locations must be writable only by trusted principals.

Optional: the `warmup_bench.rs` "not from a lock" comment predates the fingerprint mutexes on the key path; and worth a line in the PR description noting the moved `device_context` cache APIs and the `get_hash_string` -> `display_hash` rename.

goog00:
Thanks — addressed all of these. 
I moved the disk-cache tests to `warmup_suite`, added it to the GPU test script, hardened existing default cache roots with permission tests, and clarified integrity vs. authenticity in the docs. I also updated the benchmark comments and PR description for the API moves/rename.

elibol:
One correction to my earlier guidance: The 0755-retention carve-out I suggested was wrong, and the current code inherits that:

1. A traversable pre-existing root (0755/0750) can contain group/world-writable interior objects (a 0777 shard, a 0666 entry) which another user could have populated while the root was traversable, or can keep writing through an fd held open across the tightening. Tightening only the root doesn't make retained contents trustworthy. Reproduced locally: root 0755 + shard 0777 + planted entry -> `prepare_default_cache_root` returns Ok and keeps everything. Please retain a pre-existing root only when `mode & 0o077 == 0`, and reject everything else with the existing delete-and-retry error. That also simplifies the function. Flip the retention test to start from 0700 and add 0755/0750 rejection cases.

2. The JitStore security doc says `default_location` uses a per-user 0700 directory, but hardening is Unix-only. Can you scope the claim (Unix enforces 0700; Windows relies on %LOCALAPPDATA%'s default ACLs) and restore the Windows path in default_location's docs? The last revision dropped this.

3. Two book softenings: the 2 GiB bound is a soft target, not a hard cap (eviction is probabilistic and non-blocking, so transient overshoot is possible); and "anything that changes the cubin changes the key" should be narrowed to the represented inputs — two patched tileiras binaries reporting the same --version are indistinguishable, as the fingerprint docs already note.

Sorry for the churn on 1. That was my suggestion and it was wrong. 0o077 is also what sshd's StrictModes enforces, so this puts us on the standard pattern.

goog00:
Thanks for the correction. No worries about the churn, and thank you very much for the careful review.

I've addressed all three points:

* Pre-existing Unix cache roots are now retained only when `mode & 0o077 == 0`. The retention test now starts from `0700`, and rejection cases for `0755` and `0750` are covered.
* The security docs now scope `0700` enforcement to Unix, document Windows' reliance on `%LOCALAPPDATA%` ACLs, and restore the default Windows cache path.
* The book now describes 2 GiB as a soft target and limits the cache-key guarantee to represented inputs, including the same-version `tileiras` limitation. I also updated the PR description to reflect this.

`cargo fmt --all -- --check` passes.

elibol:
Excellent job with this @goog00! I _love_ this API. Very excited to see what the community builds with it.

And thanks again for your patience and persistence while we worked through all the challenges that came up along the way. Cheers!

PR #195, opened 2026-07-22
perf: bring async launch latency to parity with sync

elibol:
Closes the async-vs-sync launch latency gap: Async was ~22us at N=1, and it's now within noise of sync at every pipeline length.

The completion path no longer pays a per-launch host-callback thread hop. Short waits resolve with a bounded cuStreamQuery spin on first poll, and longer ones go through a flag-write reactor (cuStreamWriteValue32 into pinned host memory, scanned by one reactor thread) behind the `reactor` feature. Its lock-free slot protocol is model-checked under loom and miri and stress-tested under ThreadSanitizer, all in CI. Also pools streams instead of destroying them, which fixes an intermittent cuStreamDestroy segfault under concurrent teardown.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -28,7 +28,7 @@ jobs:
         run: git fetch origin main --tags
 
       - name: Set up Python
-        uses: actions/setup-python@v7
+        uses: actions/setup-python@v6
         with:
           python-version: "3.13"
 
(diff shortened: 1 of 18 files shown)


elibol:
@cryos some changes here to CI. Do you have any input?

cryos (approved):
This looks good to me.

PR #201, opened 2026-07-31
feat: host conversions for the narrow float storage types

honeyspoon:
## Summary

Add host-side conversions for the narrow floating-point storage types in
`cuda-core`:

- `f8e4m3fn`, `f8e5m2`, and `f4e2m1fn`: lossless widening to `f32` and
  round-to-nearest-even narrowing from `f32`.
- `f8e8m0fnu`: lossless widening, exponent inspection, and selection of the
  smallest power-of-two scale that covers a magnitude.
- Lossless `From<narrow> for f32` implementations, matching the convention
  used by the `half` crate.

These types already flow through `DTypeId` and Tile IR, but previously callers
had to implement their own host encoding when preparing quantized tensors or
checking kernel output.

## Conversion behavior

The scalar conversions match CUDA 13.3 host intrinsics:

| Format | Overflow | NaN input |
| --- | --- | --- |
| E4M3FN | Saturates to `+/-448` | Canonical `0x7F` |
| E5M2 | Converts to `+/-infinity` | Canonical `0x7F` |
| E2M1FN | Saturates to `+/-6` | Positive max, `0x07` |

Finite values use round-to-nearest, ties-to-even. Signed zero and gradual
underflow are preserved where the destination format supports them.

E8M0 is handled separately because it represents an unsigned power-of-two
scale rather than a signed floating-point value. `scale_covering` rounds toward
positive infinity so that scaling a block cannot make its largest element
overflow solely because the scale was rounded down. Exact powers of two remain
exact. Values outside the finite E8M0 range clamp to `2^-127` or `2^127`; NaN
maps to the reserved `0xFF` encoding.

## Implementation

E4M3FN, E5M2, and E2M1FN share a private format descriptor and one encode/decode
implementation. The public API remains inherent methods rather than adding a
new public trait:

```rust
fn to_f32(self) -> f32;
fn from_f32(value: f32) -> Self;

// f8e8m0fnu
fn exponent(self) -> Option<i32>;
fn scale_covering(magnitude: f32) -> Self;
```

The conversion path is dependency-free and marked for inlining so scalar calls
can be optimized inside host quantization loops.

## Validation

- Exhaustive decode/encode round trips for every finite E4M3FN, E5M2, and
  E2M1FN bit pattern.
- Exhaustive decode checks for all 255 non-NaN E8M0 encodings.
- Every finite rounding boundary, including ties, subnormal transitions, and
  E5M2 overflow.
- E8M0 coverage, minimality, and monotonicity across every `f32` exponent class
  and representative mantissa boundaries.
- Signed zeros, infinities, smallest/largest subnormals, and positive/negative
  signaling and quiet NaNs with minimal and maximal payloads.
- Byte-for-byte comparison against CUDA 13.3 conversions for every narrow
  decode pattern, the special-value corpus above, and 1,000,000 deterministic
  `f32` inputs.
- `cargo fmt`, scoped Clippy with warnings denied, rustdoc with warnings denied,
  Rust 1.89 MSRV, Miri, the complete CPU test suite, and workspace compilation.

## Scope

This PR only adds scalar host conversions. Packed FP4 slice conversion and
block-wise MX/NVFP4 quantization require separate layout and API decisions and
are intentionally left for follow-up work.


elibol:
@honeyspoon sorry for the delay. Mind having a look at these:
- `f8e8m0fnu(0).to_f32()` panics in debug builds and returns `0.0` in release (must be `2^-127`). Can you fix and test byte 0?
- Doc caveat on `scale_covering` for magnitudes above `2^127` (clamp means the covering property does not hold there).

honeyspoon:
updated

PR #202, opened 2026-08-01
feat(cuda-core): CUDA VMM wrappers and NVLink switch multicast

nihalpasham:
## Summary

Ports the CUDA Virtual Memory Management wrappers and the NVLink SHARP (NVLS) multicast API from cuda-oxide's cuda-core, which is migrating onto this crate.

Two commits:

1. **VMM foundation** (from cuda-oxide): RAII wrappers for `cuMemCreate`/`cuMemAddressReserve`/`cuMemMap` plus `set_access` and granularity helpers. VMM separates physical allocation, VA reservation, and mapping so physical memory from one device can be mapped into another device's address space -- the foundation for P2P symmetric heaps.
2. **Multicast wrappers** (authored by @drbh in NVlabs/cuda-oxide#606, validated end to end on an 8x B200 HGX system there): `MulticastObject` builds an NVLS team, `bind_mem` attaches per-GPU physical memory, and `Mapping::new_multicast` maps the team so `multimem.ld_reduce`/`multimem.st` operate on all bound copies at once (the mechanism behind NCCL's NVLS algorithm). Support/granularity queries allow probe-and-skip on non-NVLS systems, including drivers that predate the multicast attribute.

No bindings changes needed: the dyn-load table already binds `^cu.*` from the host `cuda.h`, and this crate's CUDA 13.2+ floor always includes the multicast API.

## Testing

- `cargo test -p cuda-core --test vmm`: full single-GPU lifecycle including a data roundtrip through the mapping and the required teardown order (passes on an RTX 5090).
- `cargo test -p cuda-core --test vmm_multicast`: two-GPU team plumbing; skips gracefully off NVLS hardware (verified skip on 2x RTX 5090; the full path ran on 8x B200 in cuda-oxide).
- `cargo fmt -- --check`, `cargo clippy`, `cargo test --no-run`, and `./scripts/run_cpu_tests.sh` all pass.


elibol:
Thanks! This looks good. It's the plumbing we need for P2P symmetric heaps and NVLS work. Verified locally: The single-GPU roundtrip passes on an RTX 5090 and the multicast test skips cleanly with one GPU.

Some requests:
1. Use `cuda_bindings::set_mem_location_id` instead of the offset-4 pointer write in `set_mem_location_device`. It already exists for exactly this and handles both bindgen layouts, so the unsafe block and the layout assumption both go away.
2. Add `vmm` and `vmm_multicast` to the target list in `scripts/run_gpu_tests.sh`.
3. Could you add DCO sign-offs (`git commit -s`, per `CONTRIBUTING.md`)? Also the multicast commit's Co-authored-by credits you rather than @drbh. I assume that's just a paste slip.
4. `align_size` assumes power-of-two granularity and can overflow for huge sizes. `debug_assert!` (`granularity.is_power_of_two()`), or `size.div_ceil(granularity) * granularity`.

One doc suggestion: The "must be dropped before" phrasing is stronger than what the driver enforces. VMM handles are refcounted and `cuMemRelease` defers the free while mappings exist, so out-of-order drops error-and-leak rather than corrupt. Worth stating that, since it's the reason these constructors are sound as safe fns. Longer term the multicast lifecycle (new -> add x N -> bind -> map) would make a nice type-state API.

nihalpasham:
Thanks, I pushed the review updates:

- Replaced the offset-based `CUmemLocation` write with `cuda_bindings::set_mem_location_id`.
- Made `align_size` handle any nonzero granularity and detect overflow, with edge-case tests.
- Added `vmm` and `vmm_multicast` to the GPU runner.
- Clarified the lifecycle docs: physical allocation handles are reference-counted, while mappings should still be dropped before VA reservations for leak-free cleanup.
- Rebased onto current main and restored the DCO trailers. The full CPU/GPU scripts and all hosted checks pass.

**Clarification on the credit:** @drbh is David Holtz, and Git already records David as the primary author of the multicast commit. I am listed as a co-author for the cutile port and follow-up adjustments. I kept that provenance, restored David's original sign-off, and added mine, so GitHub credits David as the primary author and me as co-author.

PR #204, opened 2026-08-04
fix: adapt CUDA driver flag types

Heltion:
## Summary

Adapt the six CUDA Driver API flag arguments in `cuda-core` to the unsigned
types generated by CUDA 13.3 headers.

## Root cause

`cuda-core` exposes the flag aliases as `i32`, while bindgen generates the
corresponding CUDA driver parameters as `c_uint` (`u32`). This prevents the
crate from compiling with the current CUDA 13.3 SDK on the MSVC target.

## Validation

Ran on Windows with CUDA Toolkit 13.3 and the x64 MSVC developer environment:

```text
cargo +nightly-x86_64-pc-windows-msvc check -p cuda-core --target x86_64-pc-windows-msvc
```

The check completed successfully. A 64x64 cuTile GEMM using this patch also
executed on the GPU and passed host-side element verification.

Also ran the repository-required full suite in WSL with CUDA 13.3:

```text
bash scripts/run_all.sh
```

All CPU tests, GPU tests, 23 GPU examples, and 7 benchmark smoke tests passed.

### DIFF
--- a/cuda-async/src/device_future.rs
+++ b/cuda-async/src/device_future.rs
@@ -121,13 +121,14 @@ impl<T: Send, DO: DeviceOp<Output = T>> DeviceFuture<T, DO> {
         fn host_task_sync_mode() -> Option<::core::ffi::c_uint> {
             static MODE: std::sync::OnceLock<Option<::core::ffi::c_uint>> =
                 std::sync::OnceLock::new();
+            // CUDA driver flag bindings have platform-dependent integer types, so FFI calls cast them as `_`.
             *MODE.get_or_init(
                 || match std::env::var("CUDA_ASYNC_HOST_SYNC").ok()?.as_str() {
                     "spin" | "spinwait" => {
-                        Some(cuda_bindings::CUhostTaskSyncMode_enum_CU_HOST_TASK_SPINWAIT)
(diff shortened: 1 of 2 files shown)


Heltion:
@elibol Could you please take a look when you have a chance? This fixes CUDA 13.3 MSVC bindgen flag-type mismatches. The full CPU/GPU test suite, all 23 GPU examples, and all 7 benchmark smoke tests passed in WSL with CUDA 13.3.

elibol:
Thanks for the contribution @Heltion! Taking a closer look now but LGTM so far.

elibol:
Some requests:
1. Use `as _` at all sites (including the two in `device_future.rs`). It adapts to the callee's type per platform, and clippy doesn't flag it (`as u32` adds 6 `unnecessary_cast` warnings on Linux).
2. Add a one-line module comment on why the casts exist so that it's not mistakenly modified in the future.

Heltion:
Thanks! Addressed both requests in c6b2aea: all affected flag casts now use `as _`, and I added comments explaining the platform-dependent bindgen types.

elibol:
LGTM.

The branch also edits `.github/workflows/codeql.yml` and `pages.yml`. What are those needed for? Just want to understand how they relate to the fix before merging.

Heltion:
Thanks! I may be overlooking something, but I don’t see any workflow YAML changes in this branch.

PR #206, opened 2026-08-06
feat: add num_worker_warps_per_cta hint support

EmmettZhang2020:
## Background
  `num_worker_warps_per_cta` is an optimization hint officially supported by Tile IR bytecode 13.3. It suggests how many worker warps `tileiras` should allocate to each CTA. 

## Changes

  Add `num_worker_warps_per_cta` optimization hint support to cutile-rs.

  - Supports entry-level, SM-specific hints.
  - Supports runtime overrides through `CompileOptions`.
  - Requires **Tile IR bytecode 13.3** or newer.
  - Add `num_worker_warps_per_cta` to entry optimization hints and `CompileOptions`.
  - Add tests and documentation.

## Example

  ```rust
  #[cutile::entry(optimization_hints = (
      sm_120 = (num_worker_warps_per_cta = 4,),
  ))]
  fn kernel(...) {
      // ...
  }

  let options = CompileOptions::default()
      .num_worker_warps_per_cta(8);

  kernel(...)
      .compile_options(options);
```

## Validation

  - Ran `bash scripts/run_all.sh` locally.
  - Successfully generated the documentation locally with Sphinx.

### DIFF
--- a/cutile-book/guide/jit-compilation.md
+++ b/cutile-book/guide/jit-compilation.md
@@ -111,6 +111,7 @@ Entry-level optimization hints are written on the `#[cutile::entry]` attribute:
             num_cta_in_cga = 2,
             occupancy = 2,
             max_divisibility = 16,
+            num_worker_warps_per_cta = 4, // Requires bytecode version 13.3 or newer.
         ),
         sm_90 = (
             num_cta_in_cga = 1,
@@ -127,7 +128,8 @@ use cutile::tile_kernel::CompileOptions;
 
 let opts = CompileOptions::default()
     .occupancy(4)
-    .num_cta_in_cga(2);
+    .num_cta_in_cga(2)
+    .num_worker_warps_per_cta(4); // Requires bytecode version 13.3 or newer.
 
 let _ = optimized_kernel(args)
     .compile_options(opts)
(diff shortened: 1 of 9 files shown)


elibol:
Thanks @EmmettZhang2020! This looks really straight forward. Please resolve conflicts and I'll review.

elibol:
Tile IR constrains this hint to powers of 2 in [1, 32], and the test looks like it is using some values outside of this constraint. What will `tileiras` do for hints which fall outside of this constraint? Can you note the valid range in the book/host-api docs next to the 13.3 requirement?

I noticed we don't have _any_ hint validation. I'm not asking you to write the validation part. I'm tracking that work in #208.

EmmettZhang2020:
> Tile IR constrains this hint to powers of 2 in [1, 32], and the test looks like it is using some values outside of this constraint. What will `tileiras` do for hints which fall outside of this constraint? Can you note the valid range in the book/host-api docs next to the 13.3 requirement?
> 
> I noticed we don't have _any_ hint validation. I'm not asking you to write the validation part. I'm tracking that work in #208.

Thanks for the review. What tileiras will do depends on the `num_worker_warps_per_cta` value it receives:
If the value is in the range [1, 32]:

- If it's a power of 2, the hint works.
- If it's not a power of 2, it is silently ignored and the default value of 4 is used instead. 

If the value is < 1: the default value of 4 is used silently.
If the value is > 32:
- If it's a power of 2, it is clamped to 32.
- If it's not a power of 2, it is silently ignored.

If `-Wunsupported-hints` is set, the compiler will give an explicit warning if it received invalid value.

elibol:
Merged, thanks @EmmettZhang2020! The tileiras behavior details you dug up were folded into #208, which will add the value validation across all hints.

PR #218, opened 2026-08-13
feat(jit): expose L1 and L2 cache key helpers

### LINKED ISSUE #207: feat(jit): one-liner helpers for computing L1 and L2 cache keys
The JIT has two cache keys: the in-memory `TileFunctionKey` (launch-site inputs) and the on-disk 64-hex SHA-256 (`l2_key`: bytecode + target + opt level + tileiras fingerprint). Both are computable from public pieces today, but neither is a one-liner — the L2 key in particular takes a `KernelCompiler` invocation plus manual assembly of `l2_key(...)` arguments, and the `BytecodeVersion` argument isn't obviously reachable from the compile-only artifacts. Add:

- `KernelCompiler::cache_key()` (or `jit_cache::key_for(...)`) returning the 64-hex L2 key for a described specialization — enables CI cache seeding, targeted `store.delete`, and prewarm tooling without log scraping (`CUTILE_JIT_LOG=1` is the only route today).
- An equivalent one-liner for the L1 key (today only `display_hash()` exists, as a debug label).
- L1 -> L2 conversion: the mapping exists implicitly (an in-memory miss computes the L2 key by running the frontend and serializing the bytecode) but there is no way to ask for it directly. If the two helpers above share a path, expose it — given an L1-shaped specialization, return the L2 key it would look up. Note this cannot be a pure function: it inherently runs the compiler frontend (the L2 key hashes the frontend's output), which is fine for tooling (~40-60 ms).

Refs: #193 (cache), #205 (warming/stampede — pre-seeding via these helpers is the natural companion).

goog00:
## Summary

Fixes #207.

Adds one-line helpers for deriving cuTile JIT cache keys:

- `KernelCompiler::l2_cache_key()` for compile-only specializations
- `launcher.l1_cache_key()` for the full in-memory `TileFunctionKey`
- `launcher.l2_cache_key()` for the persistent 64-hex SHA-256 key
- `launcher.specialize()` for resolving a launch specialization
- `Specialization::l2_cache_key()` for explicit L1 → L2 conversion

## Examples

Compute an L2 key for a compile-only specialization:

```rust
let l2_key = KernelCompiler::new(
    kernels::__module_ast_self,
    "kernels",
    "vector_add",
)
.generics(vec!["f32".into(), "128".into()])
.strides(&[("z", &[1]), ("x", &[1]), ("y", &[1])])
.target("sm_120")
.l2_cache_key()?;
```

Get a single key directly from a launcher:

```rust
let l1_key = kernels::vector_add(z, x, y)
    .generics(vec!["f32".into(), "128".into()])
    .l1_cache_key()?;
```

Alternatively, get the L2 key directly:

```rust
let l2_key = kernels::vector_add(z, x, y)
    .generics(vec!["f32".into(), "128".into()])
    .l2_cache_key()?;
```

Get both keys from one resolved launch specialization:

```rust
let specialization = kernels::vector_add(z, x, y)
    .generics(vec!["f32".into(), "128".into()])
    .specialize()?;

let l1_key = specialization.l1_cache_key().clone();
let l2_key = specialization.l2_cache_key()?;
```

`specialize()` resolves the specialization identity for the launch without compiling or launching the kernel.

`Specialization::l1_cache_key()` returns a reference to an already resolved `TileFunctionKey`, so it cannot fail. `specialize()` and `l2_cache_key()` return `Result` because argument materialization and frontend compilation may fail.

## Implementation

L2 derivation runs the compiler frontend, serializes Tile IR bytecode, and computes the key from the bytecode version, target, default optimization level, and current `tileiras` fingerprint.

The helpers and the real JIT disk-cache lookup share the same bytecode-to-key path:

```text
launcher.l2_cache_key()
    ▼
Specialization::l2_cache_key()
    ▼
KernelCompiler::l2_cache_key()
    ├── compiler frontend
    ├── Tile IR bytecode serialization
    ▼
current_l2_key_for_bytecode()
    ▼
l2_key(...)

compile_bytecode_cached()
    ▼
current_l2_key_for_bytecode()
    ▼
l2_key(...)
```

This ensures that the helpers return the exact key that the JIT would query.

L1 derivation does not construct the module AST or run the compiler frontend. L2 derivation runs the frontend and serializes bytecode, but does not read or write a `JitStore`, run the `tileiras` backend to produce a cubin, load a CUDA module, or populate the in-memory L1 cache.

## Testing

Validated on a CUDA 13.3 GPU server:

```bash
cargo test -p cutile --test kernel_compiler -- --nocapture

cargo test -p cutile --test warmup_suite \
  warmup::generated_cache_key_helpers_match_runtime_lookups \
  -- --exact --nocapture --test-threads=1
```
Tests verify L1/L2 parity with normal runtime lookups, key format, target sensitivity, and the absence of store or backend side effects.


elibol:
I tested this on my local GPU and looks good. I think this is a better solution than #221. 

Two naming requests before merge, since this is public API:

1. `cache_specialization()` / `cache_specialization_on()` -> `specialize()` / `specialize_on()`. "Specialize" is numba's exact verb for resolving a CUDA kernel's argument-typed instance, and it gives us stable footing toward matching numba semantics if this handle later grows compile/warmup capabilities. Please add one doc sentence making today's contract explicit. Indicate that it resolves the specialization identity for this launch, but doesn't compile or launch.
2. `L1Specialization` -> `Specialization`. The type carries both `l1_cache_key()` and `l2_cache_key()`.

With those, LGTM. I'll close #221 as superseded.

goog00:
Thank you for the review! 
I’ve applied the requested API renames and updated the PR description accordingly.

PR #225, opened 2026-08-20
fix(cutile): serialize integer division semantics

lucifer1004:
## Summary

`divi` with `Zero` rounding omitted the `rounding` attribute entirely, relying
on the Tile IR default, and neither `divi` nor `remi` serialized the
`signedness` named attribute. This makes integer division/remainder semantics
explicit in the emitted bytecode:

- `Zero` rounding now serializes `rounding=#cuda_tile.rounding<zero>`
  explicitly instead of falling back to the default;
- `divi` and `remi` now serialize `signedness=inferred_signedness`.

## Test

Adds `named_integer_arithmetic_serializes_bytecode`, which compiles a kernel
using named integer arithmetic and asserts the Tile IR bytecode is produced.
Verified on sm_120 with tileiras 13.3.36: `cargo test -p cutile --test
integer_ops` passes (7/7).


elibol:
@lucifer1004 Thanks for contributing! I'm getting around to reviewing your PRs.

PR #226, opened 2026-08-20
fix(compiler): classify scalar pointer tiles structurally

lucifer1004:
## Summary

`TileRustType::from_scalar_ptr` classified scalar pointer tiles as
`Kind::PrimitiveType`. A scalar pointer tile (`PointerTile<*mut T, {[]}>`)
lowers to a structured `!cuda_tile.tile` type like every other tile, so the
primitive classification misroutes it in code paths that treat primitives
specially (e.g. ABI handling that assumes scalars map to plain LLVM values).

Classify it structurally as `Kind::StructuredType` instead.

## Test

Adds a unit test asserting the structural classification. `cargo test -p
cutile-compiler --lib` passes.


elibol:
Thanks @lucifer1004 — you did find a real panic here, but I don't want to fix it by moving the classification.

`Kind` is a DSL-level thing for us, not a lowering fact. Everything lowers to `!cuda_tile.tile` — scalars included — so "it lowers to a tile" isn't the signal for `StructuredType`. Scalars and pointers are both `PrimitiveType` on purpose, and I'd like to keep them that way.

The actual bug is in the primitive element-type path: `get_type_ident` on `PointerTile<*mut T, {[]}>` returns the outer segment (`"PointerTile"`), which isn't a registered `ElementType`, so `get_cuda_tile_element_type_primitive` asserts and blows up. The structured path only survives because it walks the generic args and has a `Type::Ptr` arm that pulls out the pointee. So the fix belongs there — teach the primitive extractor to see through a scalar pointer tile to its pointee, the same way the structured path already does — and leave `from_scalar_ptr` at `PrimitiveType`.

Reclassifying isn't a no-op either: `Kind` gets dispatched in a bunch of places (binary-op operands are required to be `PrimitiveType`, const-value contexts too), so I'd rather not move it just to dodge the extractor gap.

If you can move the fix into the extractor and add a test that actually reproduces the panic — compile a kernel that queries the element type of a scalar pointer tile, not just an assert on the `Kind` — I'll happily take it. Thanks for digging it up.

PR #233, opened 2026-08-20
Add cuda-oxide host surface to host-side crates

elibol:
cuda-oxide's cuda-host and cuda-macros consume DeviceBuffer, the launch contract types, the embedded module, and the async launch builders. This PR copies those surfaces in as simt modules in cuda-core and cuda-async, as-is rather than reconciled, so cuda-oxide can build against the shared crates before the repository migration.

One collision on each side: LaunchConfig stays reachable only as cuda_core::simt::LaunchConfig, and cuda-async re-exports nothing at the root since the two stacks fork the same ancestor and share most type names. DriverError, IntoResult, init, and launch_kernel were already identical and are unified rather than duplicated. cuda-bindings gains the one missing helper, cu_event_elapsed_time. The simt modules are deletable as a unit at reconciliation.

Departures from the cuda-oxide originals are recorded in the module docs: the nightly-only DeviceCopy impl for primitive f16 is dropped (half::f16 remains), the DeviceCopy derive stays with cuda-oxide's cuda-macros, vmm and peer are not copied (no shared consumer; VMM arrives via #202), and oxide-artifacts is vendored as simt::artifacts since it is not on crates.io.

Full workspace compiles, 508 tests pass including cuda-oxide's launch-contract trybuild fixtures, no new clippy or rustdoc warnings. Ships in 0.3.1, not 0.3.0. Companion PR: NVlabs/cuda-oxide#1102.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -84,6 +84,39 @@ jobs:
           source "${HOME}/.cargo/env"
           ./scripts/run_cpu_tests.sh
 
+  # The shared host-side crates (cuda-bindings, cuda-core, cuda-async)
+  # support CUDA 13.0+, while the Tile stack requires 13.2+. This lane keeps
+  # the 13.0 floor honest: compile-only, no GPU, no TileIR toolchain (13.0
+  # images have no tileiras, which is exactly the point).
+  shared-crates-cuda13:
+    runs-on: linux-amd64-cpu16
+    container:
+      image: nvidia/cuda:13.0.0-devel-ubuntu24.04
+      env:
+        DEBIAN_FRONTEND: noninteractive
+    env:
+      CUDA_TOOLKIT_PATH: /usr/local/cuda
+    timeout-minutes: 30
+    steps:
+      - name: Install basic dependencies
+        run: |
+          apt update
+          apt install -y --no-install-recommends build-essential curl \
+            ca-certificates git cmake pkg-config libclang-dev zlib1g-dev \
+            libzstd-dev
+
+      - name: Clone repository
+        uses: actions/checkout@v7
+
+      - name: Install Rust
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+
+      - name: Check shared crates against CUDA 13.0
+        run: |
+          source "${HOME}/.cargo/env"
+          cargo check -p cuda-bindings -p cuda-core -p cuda-async --all-targets
+
   # Model-checks / sanitizers for the cuda-async completion reactor's lock-free
   # slot protocol. CPU-only, no GPU: loom explores the interleavings and weak-
   # memory outcomes, miri checks the UnsafeCell/pointer UB, ThreadSanitizer
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -41,7 +41,7 @@ version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7eb93bbb63b9c227414f6eb3a0adfddca591a8ce1e9b60661bb08969b87e340b"
 dependencies = [
- "object",
+ "object 0.37.3",
 ]
 
 [[package]]
@@ -424,6 +424,7 @@ dependencies = [
  "half",
  "loom",
  "once_cell",
+ "rustc-hash 2.1.2",
  "thiserror 1.0.69",
 ]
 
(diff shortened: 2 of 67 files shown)


nihalpasham:
Thanks for putting the namespaced bridge together. I’ve repaired the Tile/SIMT interop mistake on the companion PR: [cuda-oxide#1102 commit `a5096ac6`](https://github.com/NVlabs/cuda-oxide/pull/1102/commits/a5096ac6e80ffeb53d01fabcd1701bf649158f40). `cutile_inter_kernel` now checks and runs successfully on an RTX 5090.

The overall approach looks workable to me. Before we merge the cuda-oxide side and call it capability-preserving, could we carry or explicitly resolve these remaining items in this PR / the 0.3.1 release?

- **Peer access:** add the public `can_access_peer`, `enable_peer_access`, and `disable_peer_access` wrappers, plus the existing two-GPU `vmm_p2p` test. [#202](https://github.com/NVlabs/cutile-rs/pull/202) is still open and adds VMM, but not peer access.
- **VMM coverage:** finish #202 and make its VMM/multicast tests run in the GPU test script. This PR now carries `simt::vmm`, so the PR body/module docs saying VMM was not copied also need updating.
- **`DeviceCopy` compatibility:** preserve the current derive re-export (or provide an agreed source-compatible replacement) and keep the derive/trybuild tests.
- **Primitive `f16`:** preserve or explicitly decide the default behavior; `half::f16` remains, but primitive `f16: DeviceCopy` is now opt-in.
- **CUDA support:** preserve cuda-oxide’s current CUDA 12.x/13.1 and toolkit-layout support, or make CUDA 13.2+ an explicit compatibility decision rather than an incidental migration change.
- **Other public boundaries:** add compatibility tests for moved macro/module paths and `simt::artifacts`; also decide the bindings function-pointer ABI difference—ordinary calls work, but existing `unsafe extern "C" fn` assignments do not.
- **Release/test ownership:** execute the copied core/async and VMM/P2P suites in shared-crate CI, verify the 0.3.1 tarballs in bindings → core → async order, then pin that exact release in #1102.

The simple gate I’m using is:

```text
upstream parity resolved
        + #1102 216/216 and green
        + single-/two-GPU runtime tests
        + verified 0.3.1 packages
        = ready to merge
```

That keeps this a shared-package migration without giving up existing cuda-oxide behavior.

nihalpasham:
Thanks, I went through this again alongside #1102. I updated the #1102 description to be clearer and more precise, and fixed its incorrect `cutile_inter_kernel` edit. The cuda-oxide hosted checks are now green, including all 216 compile cases, and the interop example compiles and runs on an RTX 5090.

The namespaced bridge looks workable to me. Before the shared `0.3.1` release, could you please add or update the following here?

- Add the public peer-access wrappers — `can_access_peer`, `enable_peer_access`, and `disable_peer_access` — and carry the existing two-GPU `vmm_p2p` test.
- Preserve the existing `cuda_core::DeviceCopy` derive API through a publishable solution and carry its derive/trybuild tests.
- Replace the copied `simt::artifacts` implementation with the published [`oxide-artifacts 0.2.1`](https://crates.io/crates/oxide-artifacts/0.2.1) dependency and re-export it. I have published that crate from cuda-oxide. This should make both paths name the same Rust type:

  ```text
  oxide_artifacts::OwnedArtifactBundle
                =
  cuda_core::simt::artifacts::OwnedArtifactBundle
  ```

- Preserve the bindings C function type and add a compile regression. Ordinary calls work; this existing use is what currently changes type:

  ```rust
  let _: unsafe extern "C" fn(u32) -> CUresult = cuda_bindings::cuInit;
  ```

- Explicitly agree on the CTK baseline: either retain cuda-oxide's CUDA 12.x/13.1 and toolkit-discovery support, or deliberately move the shared crates to CUDA 13.2+. I mainly want to avoid changing that implicitly as part of the migration.

I am fine with primitive `f16` remaining behind the nightly feature; I no longer consider that a merge gate. The PR description could also be refreshed because the current head now includes SIMT VMM and feature-gated primitive `f16`.

The current checks are green. For the final gate, I would still like the copied core/async tests to execute rather than only compile, plus the real two-GPU P2P path and verified release packages. The release order is:

```text
oxide-artifacts 0.2.1 + cuda-bindings 0.3.1
                        ↓
                  cuda-core 0.3.1
                        ↓
                 cuda-async 0.3.1
```

If you can take care of those upstream pieces and we agree on CTK support, then with #1102 green and the final GPU/package/DCO checks passing, I think we can merge both PRs.

Issue #210, opened 2026-08-10
Pass device debug, sanitizer, and opt level flags through to tileiras


elibol:
`run_tileiras` currently invokes the backend with a fixed argument set (`--gpu-name`, `--opt-level`, `-o`). The devtools teams need three more knobs passed through for debugger, sanitizer, and profiler workflows:

1. `--device-debug` / `-g` for cuda-gdb (release code is currently close to unusable in the debugger).
2. `--sanitize=memcheck` for compute-sanitizer instrumentation.
3. Control over `--opt-level` (already sent, but not configurable; debugging likely wants level 0).

Proposed surface: fields on `CompileOptions` (they must join the JIT cache key so a debug build never serves a release cubin from cache or vice versa; opt level is already in the key, the new flags need to join it) or an env toggle for the debug case, consistent with the existing `CUTILE_JIT_*` diagnostics variables. Worth deciding whether `--device-debug` should imply opt level 0.

Requested by the debugger and sanitizer teams evaluating cutile-rs devtools readiness.

PR #237, opened 2026-08-23
feat: model full debug and add TileIR JIT overrides

almightychang:
## Summary

This PR has been rewritten on top of `main` after #236 landed with most of the original implementation. The remaining scope is:

- model TileIR optimization as `Optimization::Level(u8) | Optimization::FullDebug`
- add process-wide environment overrides for optimization, memcheck, and line information

## Why the optimization model comes first

On current `main`, `device_debug=true` plus an explicit optimized level produces an argv combination that `tileiras` rejects (`optimized debugging is currently not supported`; only O0 is accepted with `--device-debug`). Without environment overrides, both setters come from one call site, so that behavior was a reasonable API choice for the new compile options.

The override changes who produces the combination: an application can set level 3 while a wrapper independently enables debug through the environment. Neither party made an invalid local choice, but the failure would surface later during JIT in a process the wrapper does not own. Making full debug a distinct optimization state prevents that cross-party invalid combination; `FullDebug` owns both `--opt-level 0` and `--device-debug`.

Only the `opt_level` / `device_debug` pair is folded together. `lineinfo` and `sanitize_memcheck` remain independent because the measured combinations are accepted by `tileiras` (`-O3 --lineinfo`, `-O0 -g --sanitize=memcheck`, and `-O3 --sanitize=memcheck`).

This modeling change does not change cache identity. `FullDebug` still encodes opt level 0 and device-debug in flags bit 0; lineinfo remains bit 1 and memcheck bit 2. The L2 key regression test compares the new representation with the previous encoding across every valid combination. The cache domain, entry format version, and entry header layout are unchanged.

## Environment overrides

- `CUTILE_JIT_OPTIMIZATION=0|1|2|3|debug`
- `CUTILE_JIT_SANITIZE=memcheck|none`
- `CUTILE_JIT_LINEINFO=1|0` (also `true`/`false`, `yes`/`no`, and `on`/`off`)

Each variable has three states: unset preserves the code setting, an explicit value overrides it (including explicit off), and an invalid or non-Unicode value returns a clear JIT error. The values are read once per process with `OnceLock` in the existing `TileirasOptions::from_compile_options` resolution point, before the L2 cache key is constructed.

`CUTILE_JIT_LINEINFO` is included for completeness because #236 added lineinfo to the programmatic surface after the original PR was written. I am happy to remove that override if the additional knob is not wanted.

## Validation

Validated in the `cutile-ci:13.3` container used to mirror `.github/workflows/pr.yml`:

- `cargo build`
- `cargo fmt -- --check`
- `cargo clippy --all-features`
- `cargo test --no-run`
- `bash scripts/run_cpu_tests.sh`
- TileIR smoke test with `--nocapture`
- loom model check and miri slot-table checks
- `bash scripts/run_all.sh` (CPU section passed; GPU-dependent sections failed because `libcuda.so` was unavailable)

`compiler2_e2e` was confirmed to execute rather than silently skip: it printed its target and produced non-empty cubins with the real CUDA 13.3 `tileiras`. In particular, application code specifying `opt_level(3)` plus `CUTILE_JIT_OPTIMIZATION=debug` resolved to `--opt-level 0 --device-debug` and compiled successfully. The real backend also accepted the directly tested default, O0, debug, debug+memcheck, O3+memcheck, and lineinfo argv combinations.

GPU launch paths were not validated on this machine because the container had no CUDA driver (`libcuda.so`, driver error 303); those paths rely on CI with a compatible GPU and driver. The local wrapper also lacks the workflow's `seccomp=unconfined` option, so the ThreadSanitizer command stopped at `setarch: Operation not permitted`; the dedicated CI job remains the verification for that check.


elibol:
@almightychang I just merged #236 and it overlaps substantially with this impl. Sorry, I never flagged on #210 that work was already in flight and I only realized the overlap after merging and looking at this PR for the first time.

Otherwise, this PR looks good. There are some ideas here that I'd like to see in main with you as author, if you're up for it:
1. The env overrides (CUTILE_JIT_OPTIMIZATION / CUTILE_JIT_SANITIZE). Right now, nothing on main can switch a kernel to a debug or sanitizer build without a code change. Main now resolves all of these flags in one place (`TileirasOptions::from_compile_options`), so the override slice should rebase cleanly.
2. Your `Level` | `FullDebug` modeling. You measured the same thing we did (tileiras rejects --device-debug above level 0); main expresses it as "device_debug implies 0 unless set explicitly", but making the invalid combination unrepresentable is a fair argument. The API is days old, so there's still time to make the case for it.

If you're up for the env-override follow-up, it's yours. Sorry again!

almightychang:
@elibol 
Ah, I missed #236 was already in flight.
I will rebase onto main and narrow down the scope to the follow ups.

almightychang:
@elibol 
Reworked this PR on top of `main` after #236 landed. The updated branch now contains two focused commits:

1. Model optimization as `Level(u8) | FullDebug`, while preserving the existing cache-key bit layout.
2. Add process-wide JIT environment overrides and regression coverage for their composition with code-level options.

In particular, the tests now pin the critical case where code requests optimization level 3 but `CUTILE_JIT_OPTIMIZATION=debug` resolves to full debug (`--opt-level 0 --device-debug`). They also cover preserving code settings when overrides are unset, explicit on/off behavior for sanitizer and lineinfo, and invalid environment values.

Validation was run in the `cutile-ci:13.3` container: all 142 `cutile-compiler` library tests and the formatting check pass. I also verified the effective TileIRAS argv combinations, including the level-3-to-debug override case.

PR #241, opened 2026-08-24
fix: version persisted tuning cache keys

almightychang:
## How this was found

While addressing #237's review follow-ups, including preserving the cache-key bit layout, I found a second consequence of #236's L2-key change.

#236 changed `cutile-compiler/src/jit_cache.rs:749` from hashing `[opt_level]` to `[opts.opt_level, opts.flags_byte()]`. Since `put_field` length-prefixes each field, every key changes, including defaults. For one real kernel with the same target, `tileiras 13.3 V13.3.36`, and default options:

```text
before: 1483ae7fad3dbc613f5cdba7c55a816105ede32aa5c3003b81533166a9970b11
after:  bdb251ef25a9034cb6de88f6dadc0b52633df1e15a0ccf8aa826a186c1f8203b
```

#236 documented the disk-cache cost as a one-time recompile (`jit_cache.rs:802-804`). However, the same key is also persisted in `cutile::tune::Record` (#224, released in 0.3.0), where drift causes `load_verified` to refuse the record. Existing version checks cannot soften this: the `cutile_version` warning comes after key verification (`tune.rs:1007` vs. `:996`), both versions are `0.3.0`, and `RECORD_SCHEMA` remains 1.

## What this PR does

This PR stores the key encoding with each digest:

```rust
L2Key::Tagged { schema, digest } | L2Key::Untagged(String)
```

Verification now has three outcomes:

- same schema, same digest: accept;
- same schema, different digest: refuse as drift;
- different or absent schema: warn and skip an incomparable check.

The tag is per key so partially retuned records can safely contain mixed encodings. `#[serde(untagged)]` keeps wire compatibility: a bare-string key written by released 0.3.0 still parses as `Untagged`.

This keeps #236's necessary flags-aware cache identity and preserves strict verification for comparable keys, instead of reverting the key change or globally loosening the loader.

Validation in `cutile-ci:13.3` with an RTX 5080 (`--gpus all`):

- `cargo test -p cutile --features experimental-tune`: 39 test binaries, 0 failures, including both GPU tests;
- `cutile-compiler` library tests: 135 passed;
- `cargo fmt`: clean.

## Open question: `gate`

`Record::gate` says drift is warned (`tune.rs:797`), and `load_verified` lists it under **WARNS** (`tune.rs:951`), but `Workspace` has no gate field and the loader never compares it.

Should I file an issue or send a follow-up PR? The fix is either to implement the check by adding `gate` to `Workspace`, or to correct the docs—depending on whether `gate` was intended to be a verification gate.

### DIFF
--- a/cutile-book/guide/performance.md
+++ b/cutile-book/guide/performance.md
@@ -207,10 +207,10 @@ How it works:
 
 ### Committing winners
 
-A `tune::Record` persists winners so production loads them instead of re-searching. It is one JSON file per kernel, written by `save` and intended to be committed next to the code it tunes. The file holds a provenance header (kernel name, source hash, cutile version, `tileiras` fingerprint, target architecture, search-space hash) and one entry per shape-class bucket: the winning `Config`, its median, and optionally the winner's JIT cache key.
+A `tune::Record` persists winners so production loads them instead of re-searching. It is one JSON file per kernel, written by `save` and intended to be committed next to the code it tunes. The file holds a provenance header (kernel name, source hash, cutile version, `tileiras` fingerprint, target architecture, search-space hash) and one entry per shape-class bucket: the winning `Config`, its median, and optionally the winner's JIT cache key. A stored key carries the key encoding that produced it, because cutile can change that encoding without any change to the kernel or the toolchain.
 
 ```rust
-use cutile::tune::{Record, RecordEntry, Workspace};
+use cutile::tune::{L2Key, Record, RecordEntry, Workspace};
 
 // After tuning: store the winner for this bucket.
 let mut record = Record::new(&workspace);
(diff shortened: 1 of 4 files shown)


elibol:
Nice catch! Both of our review passes missed it...

The design looks good to me. One small addition: If the schema (the actual schema, not `L2_KEY_SCHEMA`, which I'll rename to `L2_KEY_VERSION` for clarity) changes without bumping `L2_KEY_SCHEMA`, a test ought to fail. I pushed a small test onto your branch that pins one exact digest, so any change to the key computation fails CI with instructions to bump `L2_KEY_SCHEMA`. Feel free to adjust or rewrite it.

Ran the full tune suite and the GPU record e2e locally, all green. LGTM once CI agrees.

PR #262, opened 2026-09-03
fix: round f8 constants to nearest even

### LINKED ISSUE #250: cutile-ir: f8 constant encoder truncates the mantissa instead of round-to-nearest-even
`convert_to_f8` in `cutile-ir/src/bytecode/encoding.rs` encodes f8 constant
literals by **truncating** the mantissa instead of rounding to nearest (ties to
even). Both f8 formats go through it (`f64_to_f8e4m3fn`, `f64_to_f8e5m2`):

- Normal path (`encoding.rs:319-321`):
  `let truncated_man = (f64_man >> (52 - man_bits)) as u8 & man_mask;` — a right
  shift that drops the low mantissa bits with no rounding.
- Subnormal path (`encoding.rs:316`): same, a shift with no rounding.

### Why it's wrong / inconsistent
- f16/bf16 constants are encoded via the `half` crate (`from_f64`), which rounds
  to nearest-even — only the hand-rolled f8 path truncates.
- The device `cvt` instruction for these formats is round-to-nearest-even
  (with satfinite).
- The host-side narrow-float converters added in #201 (`cuda-core`) round to
  nearest-even.

So an f8 constant that isn't exactly representable can encode to a **different
value** depending on which path handles it, and the bytecode-constant path
always biases toward zero.

### Impact
Silent, up to ~1 ULP at f8 precision. Constants that are exactly representable
(0, 1, powers of two, small integers) are unaffected; the gap only appears for a
literal whose exact value lands between two representable f8 values and would
round up under RNE. Block-scale constants (NVFP4/MX) are a plausible place to
hit it.

### Fix sketch
Round-to-nearest-even on the discarded bits (guard/round bit + sticky, ties to
even), with carry propagation: a mantissa carry-out increments the exponent, and
an exponent that reaches the max boundary must saturate correctly (max-finite for
E4M3FN; inf/NaN policy for E5M2). Apply to both the normal and subnormal paths.
A regression test with a value just above a midpoint (currently truncates down,
should round up) pins it.

### Provenance
Flagged by @drbh while adding the host narrow-float conversions in #201; filing
it here so it's tracked rather than lost in the PR thread.

w3lld1:
## Summary

- I changed the f8 constant encoder to use round-to-nearest-even for both normal and subnormal values.
- I propagate rounding carries into the exponent and handle the subnormal-to-normal transition.
- I preserve E4M3FN saturation and E5M2 infinity behavior at the upper boundary.
- I added regression coverage for ties, values just above a midpoint, mantissa carry, underflow, and overflow boundaries.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p cutile-ir --all-targets`
- `cargo clippy -p cutile-ir --lib` (completed with existing unrelated warnings)
- `git diff --check`

I did not run `scripts/run_all.sh` locally because this environment does not have a CUDA toolkit or GPU, so I am leaving the PR as a draft for the repository's full CUDA/GPU validation.

Fixes #250

### DIFF
--- a/cutile-ir/src/bytecode/encoding.rs
+++ b/cutile-ir/src/bytecode/encoding.rs
@@ -266,25 +266,32 @@ fn convert_to_f8(
 
     let max_exp = (1i32 << exp_bits) - 1;
     let man_mask = (1u8 << man_bits) - 1;
+    let max_finite_exp = if nan_only_all_ones {
+        max_exp
+    } else {
+        max_exp - 1
+    };
+    let max_finite_man = if nan_only_all_ones {
+        man_mask - 1
+    } else {
+        man_mask
+    };
+    let max_finite = (sign << 7) | ((max_finite_exp as u8) << man_bits) | max_finite_man;
 
     // Handle special values.
     if f64_exp == 0x7FF {
-        // Inf or NaN
-        if f64_man != 0 || (nan_only_all_ones && f64_man == 0) {
-            // NaN (or Inf mapped to NaN for formats without infinities)
-            if nan_only_all_ones {
-                return (sign << 7) | ((max_exp as u8) << man_bits) | man_mask;
-            } else {
-                // IEEE NaN: all-ones exponent, non-zero mantissa
-                return (sign << 7) | ((max_exp as u8) << man_bits) | 1;
-            }
+        if f64_man != 0 {
+            // NaN: use a non-zero mantissa (all ones for F8E4M3FN).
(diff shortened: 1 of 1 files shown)


elibol:
Thanks for this. The nearest-even rounding for f8 constants landed in #275 (in 0.4.0), including the subnormal boundaries and E4M3FN finite saturation, and that closed #250. Closing this one as superseded rather than asking you to rebase onto a fix that is already in.

PR #270, opened 2026-09-07
fix: lifetime-brand CudaGraph so captured buffers cannot dangle or race

elibol:
**Breaking API changes (cuda-async):**
- `CudaGraph<T>` → `CudaGraph<'a, T>`: the graph mutably borrows every captured buffer for its lifetime, so dropping, moving, or writing one while the graph can still replay is now a compile error (previously a use-after-free / data race reachable in safe code).
- `take_output` removed: outputs are reachable only after a completed replay, via `outputs(&mut self)`, `into_inner(self)`, or the `&mut T` returned by `replay()` / `GraphLaunch::sync`.
- `launch` takes `&mut self` and returns `GraphLaunch<'_, 'a, T>`; the old `graph.launch().sync_on(graph.stream())` idiom becomes `graph.replay()`.
- `DeviceOp::graph`/`graph_on` gained a lifetime parameter (`where Self: 'a`).

New: `replay()`, `outputs()`, `into_inner()`, and `update_with()` (refresh an input buffer the graph owns). Self-referential graph structs migrate by moving buffers into the scope closure and returning them as `T` (see the updated `cuda_graphs` example). Compile-fail UI tests pin the rejected programs; GPU tests cover sequential/async replays and `into_inner`. cutile's public API is unchanged.

### DIFF
--- a/cuda-async/src/cuda_graph.rs
+++ b/cuda-async/src/cuda_graph.rs
@@ -16,13 +16,24 @@ const CU_STREAM_CAPTURE_MODE_RELAXED: sys::CUstreamCaptureMode = 2;
 
 /// A captured and instantiated CUDA graph, ready for replay.
 ///
-/// Created via [`CudaGraph::capture`], which runs a [`DeviceOp`] once on a
-/// capture stream, recording all GPU work into a graph. The graph can then
+/// Created via [`CudaGraph::capture`] or [`CudaGraph::scope`]. The graph can
 /// be replayed any number of times via [`launch`](CudaGraph::launch).
 ///
-/// All device pointers used by the operation are baked into the graph at capture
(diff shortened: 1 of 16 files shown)


elibol:
Closing. The use-after-free this targeted (#272) was fixed at runtime by #275, which keeps recorded storage alive across replays and rejects conflicting cross-stream accesses. What remained here after rebasing was an API reshaping (`launch(&mut self)`, `outputs()` replacing `take_output`) with no safety or performance benefit to justify the breaking change. The branch stays on the internal mirror for reference; the rewritten replay-safety docs may return as a docs-only PR.

elibol:
Reopening as a documentation-only change: the replay-safety contract that #275 established (recorded storage kept alive across replays, per-replay lease reacquisition, cross-stream conflict rejection) is described in the graph docs. No API or code changes.

PR #278, opened 2026-09-13
fix(cutile): synchronize stream in to_host_vec before dropping owned tensor

### LINKED ISSUE #252: [BUG]: Possible use after free due to deallocation being queued on a different stream before `sync_on` is called
### Operating System

Ubuntu 24.04

### GPU and GPU Architecture

NVIDIA GeForce RTX 5070, sm_120

### Error Description

Compute Sanitizer reports a possible use-after free when using `tensor.to_host_vec().sync_on(&stream)`. 
This is different from the UAF reported in #242 (and fixed by #179?) since here future cancellation is not involved.

```
========= COMPUTE-SANITIZER
========= Use-after-free on allocation of size 1 bytes at 0x900000000
=========     Address 0x900000000 is potentially accessed after it is free'd
=========
=========     Saved host backtrace up to driver entry point at cudaMemcpy time
=========         Host Frame: cuMemcpyDtoHAsync_v2
=========                        in libcuda.so.1
=========         Host Frame: [...]
=========                        in cuda-core/src/cudarc_shim.rs:730
=========         Host Frame: [...]
=========                        in cuda-core/src/api.rs:131
=========         Host Frame: [...]
=========                        in cutile/src/api.rs:335
=========         Host Frame: [...]
=========                        in cutile/src/api.rs:360
=========         Host Frame: [...]
=========                        in cuda-async/src/device_operation.rs:379
=========         Host Frame: [...]
=========                        in cutile-to-host-vec-repros/src/bin/
=========                           to_host_vec_stream_ordering.rs:18
=========
=========     Saved host backtrace up to driver entry point at cudaFreeAsync time
=========         Host Frame: cuMemFreeAsync
=========                        in libcuda.so.1
=========         Host Frame: [...]
=========                        in cuda-core/src/cudarc_shim.rs:655
=========         Host Frame: [...]
=========                        in cuda-core/src/api.rs:99
=========         Host Frame: [...]
=========                        in cuda-async/src/device_buffer.rs:60
=========         Host Frame: [...]
=========                        in cuda-async/src/device_context.rs:359
=========         Host Frame: [...]
=========                        in cutile/src/api.rs:361
=========         Host Frame: [...]
=========                        in cuda-async/src/device_operation.rs:379
=========         Host Frame: [...]
=========                        in cutile-to-host-vec-repros/src/bin/
=========                           to_host_vec_stream_ordering.rs:18
=========
copy_result=ok host_byte=0
========= ERROR SUMMARY: 1 error
```

### Minimal Reproduction Example

```rust
use cutile::error::Error;
use cutile::prelude::*;

fn main() -> Result<(), Error> {
    let device = Device::new(0)?;
    let stream = device.new_stream()?;
    let tensor = api::zeros::<u8>(&[1]).sync_on(&stream)?;

    // This is ordinary safe API usage. Memcheck reports that the owned tensor
    // is freed on cuTile's deallocator stream before this stream is synchronized.
    let host = tensor.to_host_vec().sync_on(&stream)?;
    assert_eq!(host, [0]);
    println!("copy_result=ok host_byte={}", host[0]);
    Ok(())
}
```

then run

```sh
compute-sanitizer --tool memcheck --track-stream-ordered-races use-after-free
```

emersonbusson:
## Summary

Fixes #252.

Synchronizes the execution stream in `CopyDeviceToHostVec::execute` before dropping `self.tensor`, eliminating a stream-ordered race and use-after-free between the in-flight DtoH memcpy and asynchronous deallocation on the `deallocator_stream`.

## Root Cause

When copying an owned tensor to host memory via `tensor.to_host_vec().sync_on(&stream)`:
1. `CopyDeviceToHostVec` holds ownership of the source tensor via `self.tensor: Arc<Tensor<T>>`.
2. In `CopyDeviceToHostVec::execute`, `memcpy_dtoh_async` is enqueued on `ctx.get_cuda_stream()`.
3. `execute(self, &ctx)` consumes `self` by value, meaning `self` (and its inner `self.tensor`) is dropped at the return of `execute`.
4. Dropping the owned tensor triggers `DeviceBuffer::drop`, which immediately enqueues `cuMemFreeAsync` on the thread's `deallocator_stream`.
5. Because `deallocator_stream` does not establish an event wait on `ctx.get_cuda_stream()`, the free operation on `deallocator_stream` races with the in-flight copy on `ctx.get_cuda_stream()`, triggering a stream-ordered use-after-free flagged by Compute-Sanitizer (as reported in #252).

## Fix

By explicitly calling `ctx.get_cuda_stream().synchronize()` inside `CopyDeviceToHostVec::execute` before `self.tensor` is dropped:
- The device read on `ctx.get_cuda_stream()` is guaranteed to be 100% complete before any deallocation can be submitted to the `deallocator_stream`.
- Ensures the safety contract (`SAFETY: ... returns only once the copy has completed, so all size elements are initialized here`) holds deterministically across all stream configurations.

## Validation

- Added regression test `to_host_vec_owned_sync_on_stream_order` in `cutile/tests/gpu/tensor.rs` matching the reproduction snippet from #252.

### DIFF
--- a/cutile/src/api.rs
+++ b/cutile/src/api.rs
@@ -342,11 +342,14 @@ impl<T: DType> DeviceOp for CopyDeviceToHostVec<T> {
             unsafe {
                 memcpy_dtoh_async(host.as_mut_ptr(), cu_deviceptr, size, ctx.get_cuda_stream())
             }?;
+            // Synchronize the execution stream before dropping `self.tensor` (owned by `self`),
+            // ensuring the device read finishes before any asynchronous deallocation is
+            // enqueued on the deallocator stream (#252).
+            unsafe { ctx.get_cuda_stream().synchronize() }?;
         }
-        // SAFETY: `cuMemcpyDtoHAsync` into pageable host memory (a `Vec`'s
-        // heap buffer is pageable) returns only once the copy has completed,
-        // so all `size` elements are initialized here, and `size` is exactly
-        // the capacity reserved above.
+        // SAFETY: `cuMemcpyDtoHAsync` into pageable host memory followed by stream
+        // synchronization ensures the copy has completed, so all `size` elements
+        // are initialized here, and `size` is exactly the capacity reserved above.
         unsafe { host.set_len(size) };
         Ok(host)
     }
--- a/cutile/tests/gpu/tensor.rs
+++ b/cutile/tests/gpu/tensor.rs
@@ -48,3 +48,14 @@ fn from_raw_parts_rejects_shape_storage_mismatch() {
         )
     };
 }
+
+/// Regression test for #252: owned tensor `to_host_vec().sync_on(&stream)` must not race
+/// with asynchronous deallocation on the deallocator stream.
+#[test]
+fn to_host_vec_owned_sync_on_stream_order() {
+    let device = cuda_core::Device::new(0).expect("Failed to create device");
+    let stream = device.new_stream().expect("Failed to create stream");
+    let tensor = api::zeros::<u8>(&[1]).sync_on(&stream).expect("zeros failed");
+    let host = tensor.to_host_vec().sync_on(&stream).expect("to_host_vec failed");
+    assert_eq!(host, vec![0]);
+}


elibol:
Thanks for tracking this down. #275 landed first and made the same change inside `CopyDeviceToHostVec::execute`, plus a storage lease that also covers the async path, so this no longer applies to main. #252's sanitizer repro is clean on main and reproduces on the commit before #275, so I'm closing this as superseded.

If you'd like a follow-up, a sanitizer-gated regression test around that repro would be welcome; the one here checks a single byte and passes on the old code too.

PR #291, opened 2026-09-18
doc: fix data parallel MLP tutorial

notbowen:
The provided code snippet in the data parallel MLP tutorial doesn't compile as `Tensor<f32>` doesn't implement `.clone()`. The fix, as shown in `cutile-examples/examples/async_mlp.rs` is to simply wrap the `Tensor<f32>` in `Arc<...>`.

This PR implements the `Arc<...>` fix, and compiles and runs on cutile-rs 0.3.1.


elibol:
Thanks @notbowen! Much cleaner.

Issue #208, opened 2026-08-06
Validate optimization hint values against Tile IR constraints


elibol:
Optimization hints are forwarded to the backend without value validation. The Tile IR spec constrains each hint, but cutile only checks that the value is an integer, so an out of range value becomes a late tileiras verifier error instead of an early cutile error with a source location.

Constraints per the Tile IR attribute definitions:

| Hint | Constraint |
|---|---|
| `num_cta_in_cga` | power of 2 in [1, 16] |
| `occupancy` | integer in [1, 32] |
| `num_worker_warps_per_cta` | power of 2 in [1, 32] |
| `latency` (op level) | integer in [1, 10] |

We should validate at both entry points: the `optimization_hints` macro path (`SMHints` setters) and the runtime `CompileOptions` builder. Each should reject out of range values with an error that states the valid range. The book and host-api reference should document the ranges next to each hint.

Came up in #206, where the forwarding test used values the backend verifier rejects. Version compatibility is already enforced (the 13.3 gate for `num_worker_warps_per_cta`); this issue is about value constraints only.

elibol:
Constraint behavior, confirmed against the Tile IR implementation (plus the empirical report on #206): hint value checking in the backend is warning only (behind `-Wunsupported-hints`) and off by default. For out of range values the backend silently ignores the hint, substitutes its default, or clamps. For `num_worker_warps_per_cta` specifically: a non power of 2 in range is ignored (default 4 used), values below 1 get the default, powers of 2 above 32 are clamped to 32, and non powers of 2 above 32 are ignored. The only hard error is the physical warps per CTA limit, which is not a hint check.

So without cutile side validation a bad hint silently produces a different kernel configuration than requested, with no diagnostic anywhere. Validation here is load bearing, not a nicety.

Two scope additions:
1. Consider forwarding `-Wunsupported-hints` to tileiras in debug mode (e.g. when `CUTILE_JIT_LOG` is set) so backend hint diagnostics surface during development.
2. Hint checking must stay entirely on our side; do not assume the backend will ever reject a value.

0z5a:
Hi @elibol , I'd like to take this one. I'll implement value validation for the optimization hints in both the `SMHints` and `CompileOptions` paths, add coverage for valid/invalid boundary cases, and update the relevant docs with the supported ranges.

0z5a:
Implemented in #292.

[...]

PR #292, opened 2026-09-18
fix(compiler): validate optimization hint values before backend compilation

0z5a:
**Incremental review:** [Files changed](https://github.com/NVlabs/cutile-rs/pull/292/files) against `main`.

Optimization hints were forwarded to the backend with only an "is it an integer" check, so an out-of-range value surfaced as a late `tileiras` verifier error instead of an early error with a source location.

This adds one value-domain gate and routes every path that can set a hint through it.

## Accepted values

Taken from the Tile IR attribute definitions, per the issue's table:

| Hint | Accepted values | Where checked |
|---|---|---|
| `occupancy` | integer in `[1, 32]` | macro setter, builder, `CompileOptions::validate` |
| `num_cta_in_cga` | power of two in `[1, 16]` | same |
| `num_worker_warps_per_cta` | power of two in `[1, 32]` | same |
| `latency` (per-op) | integer in `[1, 10]` | load/store call-site parsing |

`max_divisibility` is deliberately **not** range-checked: it is a ceiling on auto-inferred divisibility rather than a Tile IR attribute with a fixed legal set, so constraining it here would invent a rule the backend does not have.

## Notes on the implementation

- **The entry-level diagnostic keeps a real source position.** `crate::hints` stays free of compiler dependencies, so the caller supplies a `HintLocationResolver`; the compiler passes one backed by the owning module's `SpanBase`, which resolves the offending expression to real `file:line:col`. The runtime builder has no span at all and reports the option name and value rather than inventing a position.
- **A public field cannot bypass the check.** The hint fields on `CompileOptions` are public, so the builder is not the only way to fill them. `CompileOptions::validate` re-checks every field and the compile boundary calls it. That also places the check ahead of IR generation and the JIT cache key, so a cache hit cannot admit an invalid configuration.
- **The setters now return `Result` rather than `Self`.** The issue asks for the runtime builder to reject out-of-range values too, not only the macro path. Chaining is preserved for callers that propagate with `?`; the one call site in the repository that cannot propagate, `cutile/tests/debug_info.rs` (`CompileOptions::new().occupancy(2).sanitize_memcheck(true)`), states its expectation with `.expect("2 is a valid occupancy")`.
- **`Latency::<0>` keeps its existing meaning.** The crate spells "no latency hint" as `Latency::<0>` at load/store call sites, so zero stays accepted and only an explicitly out-of-range value is rejected. The per-op `latency` domain is therefore `[1, 10]` for the values that are actually forwarded.

## Verification

| Check | Result |
|---|---|
| `cargo test -p cutile-compiler --lib hints::` | 22 passed, 0 failed |
| `cargo test -p cutile-compiler --lib` | 170 passed, 0 failed |
| `cargo test -p cutile --test optimization_hints` | 35 passed, 0 failed |
| `cargo test -p cutile --test specialization_bits` | 13 passed, 0 failed |
| `cargo test -p cutile --test warmup` | 11 passed, 0 failed |
| `cargo test -p cutile --test debug_info` | 4 passed, 4 ignored |
| `cargo test -p cutile --test gpu` (on an L20, sm_89) | **100 passed**, 0 failed, 2 ignored |
| `cargo test -p cutile` (whole package) | identical target for target on the pristine tree at this base; only `dtype_float_ops` fails (2 tests, pre-existing) |
| `cargo build --tests -p cutile-compiler -p cutile` | every test target compiles |
| `cargo fmt --all -- --check` | clean |

The GPU suite is still **100 passed** on this base: `--test gpu` reports 91 passed / 2 ignored and `--test submission_lifetimes` reports 9 passed. Main's #295 isolated the lifetime tests into their own target, so what the pre-rebase tree reported as 100 in the `gpu` target alone is now split across two.

The two failures in the `cutile` package are pre-existing on the pristine tree at this base
(`dtype_float_ops::create_ones_f8e4m3fn_and_convert_to_f32` and
`create_ones_f8e5m2_and_convert_to_f32`); they fail identically with and without this change,
target for target, so there is no new regression.

## Measured effect on the invalid path

One kernel per process (the two entries share a module, so the second one would hit the
in-process cache), `tileiras` invocations counted by a pass-through wrapper that is the only
thing on `CUTILE_TILEIRAS_PATH`, with a fresh `CUTILE_JIT_CACHE_DIR` per run, on an L20:

| | pristine tree at this base | this change |
|---|---|---|
| `occupancy = 33` (out of range) | **ACCEPTED**, 946 bytes of Tile IR, `tileiras` called **1×** | **rejected in the frontend**, `tileiras` called **0×** |
| time to the error (median of 3) | 419.7 ms (167.8 ms frontend + 251.9 ms backend) | **111.3 ms**, no backend |
| `occupancy = 1` (in range, control) | ACCEPTED, 949 bytes, `tileiras` called 1× | ACCEPTED, 949 bytes, `tileiras` called 1× |

The control matters twice over: it shows the counter is live, and it shows the change does not
touch the legal path — the bytecode is identical and the backend still runs. What changes is
only what an invalid value costs: the pristine tree pays a full backend compile before the
verifier reports what the frontend could have said, which is the cost an autotuner pays on
every rejected configuration. The backend time itself varies with host load (150–407 ms here),
so the structural fact to read is the invocation count, 1 → 0.

New tests:

- 22 unit tests in `hints.rs` covering each domain's boundaries, negative values (asserting the reported value is not an unsigned reinterpretation), the error text naming the hint and the range, the macro path's located diagnostic, unlocated diagnostics staying `Generic`, `max_divisibility` staying accepted, and a directly written field being caught by `validate` and rejected by `apply_compile_options`.
- 2 integration tests in `cutile/tests/optimization_hints.rs`: the macro path rejects three out-of-range kernels with a diagnostic naming the hint, the range, and a source location; and the legal kernels next to them still compile, so a rejection is attributable to the value rather than to the surrounding shape.

Refs #208.


0z5a:
Rebased onto `main` (`e04245b`) — the branch is now `00bb516` + `03b631f`, and the PR is
mergeable again.

Conflicts and how they were resolved:

- `cutile-compiler/src/hints.rs` — main added `DebugInfoLevel` and an explicit
  `impl Default for CompileOptions` in the same region this change edits. Both sides are kept;
  the value-domain gate, the located diagnostics and the `Result` setters are unchanged.
- `cutile-compiler/src/compiler/_function.rs` — main reworked the `CallFrame` documentation and
  added `source_location`. Git's auto-merge and an independent three-way merge produce
  byte-identical files, so nothing here is a judgement call.
- One call site needed adapting to the `Result`-returning setters: the new
  `cutile/tests/debug_info.rs` chains `.occupancy(2).sanitize_memcheck(true)`, and gets
  `.expect("2 is a valid occupancy")` between them. It is the only call site in the repository
  the change did not already cover, and it is why the diff is now 11 files instead of 10.

E2E re-run on the rebased tree (L20, sm_89, GPU 3):

| Check | Result |
|---|---|
| `cargo build --tests -p cutile-compiler -p cutile` | every test target compiles |
| `cutile-compiler --lib hints::` | 22 passed / 0 failed |
| `cutile-compiler --lib` | 170 passed / 0 failed |
| `cutile --test optimization_hints` | 35 passed / 0 failed |
| `cutile --test specialization_bits` | 13 passed / 0 failed |
| `cutile --test warmup` | 11 passed / 0 failed |
| `cutile --test debug_info` | 4 passed / 4 ignored |
| `cutile --test gpu` | **100 passed** / 0 failed / 2 ignored |
| `cargo test -p cutile` (whole package, both trees) | identical target for target; only `dtype_float_ops` fails, on the pristine tree too |
| `cargo fmt --all -- --check` | clean |

The GPU suite is still **100 passed** on this base: `--test gpu` reports 91 passed / 2 ignored and `--test submission_lifetimes` reports 9 passed. Main's #295 isolated the lifetime tests into their own target, so what the pre-rebase tree reported as 100 in the `gpu` target alone is now split across two.

Re-measured rejection cost, one kernel per process, `tileiras` invocations counted by a
pass-through wrapper, fresh JIT cache dir per run:

| `occupancy = 33` | pristine tree at this base | this change |
|---|---|---|
| outcome | ACCEPTED, 946 bytes of Tile IR | rejected in the frontend |
| `tileiras` invocations | **1** | **0** |
| time to the error (median of 3) | 419.7 ms (167.8 frontend + 251.9 backend) | **111.3 ms** |
| `occupancy = 1` control | ACCEPTED, 949 bytes, 1 invocation | ACCEPTED, 949 bytes, 1 invocation |

The control is what makes the count meaningful: the legal configuration still compiles and still
invokes the backend in both trees, so the change removes the wasted backend compile only for the
configuration that was never valid.

Issue #205, opened 2026-08-05
Track cross-process JIT compilation stampede on cold cache misses


goog00:
## Context

Follow-up to #193.

The in-process JIT cache uses `OnceCell` to compile each kernel only once per process. The persistent disk cache can be shared across processes, but compilation is not deduplicated across process boundaries.

When multiple processes sharing the same cache directory cold-start concurrently — commonly with one process per GPU — they can all miss the same key, spawn `tileiras`, and compile the same kernel.

This is correct: atomic temporary-file + rename writes ensure that the cache converges on one identical entry. However, the first parallel launch wastes N−1 compiles. Subsequent starts hit the populated cache.

## Decision for #193

Merge without a per-key cross-process compilation lock and track the limitation here.

A blocking `flock` was left out because:

- A slow or hung lock holder could block every waiter.
- Waiters cannot reliably distinguish active compilation from a stuck compiler.
- File-lock semantics and reliability can vary across NFS deployments, where shared caches are likely to live.
- The stampede affects only a cold cache and converges correctly afterward.

Users can avoid the initial stampede by warming the shared cache with one process before launching parallel workers. This workaround is documented in `cutile-book/guide/jit-compilation.md`.

## Follow-up

Revisit cross-process compilation coordination if cold-start stampedes become a significant real-world cost.

Any future solution should:

- Where cross-process coordination is supported, allow only one process to compile a missing cache entry while the others reuse its result.
- Document whether this coordination is supported on network filesystems such as NFS.
- Where coordination is unavailable, fall back to redundant compilation rather than fail a kernel launch or block indefinitely.

elibol:
Two ideas worth tracking here, based on how other systems handle persistent JIT caches:

1. A minimum compile time threshold for persisting (JAX defaults to 1 second). Only persist compiles that were expensive enough to be worth it. The frontend residual is about 40 to 60 ms and tileiras dominates a cold compile, so this keeps the cache focused on the expensive kernels. It also shrinks the cost of the cold start stampede, which is what this issue tracks.

2. Per shard eviction (Mesa moved to this in 23.1). Give each of the 256 shards a budget of capacity/256 and evict within the shard that was just written. Deterministic, no probabilistic trigger, no full tree scan. A simplification path if the current amortization ever becomes hard to maintain.

0z5a:
> Two ideas worth tracking here, based on how other systems handle persistent JIT caches:
> 
[...]

PR #294, opened 2026-09-18
bench(jit): measure cross-process cold-cache compilation stampedes

0z5a:
**Incremental review:** [Files changed](https://github.com/NVlabs/cutile-rs/pull/294/files) against `main`.

Phase 1 for #205: a reproducible harness, raw data, and conclusions for the cross-process cold-cache compilation stampede. **No coordination mechanism is added here, and the runtime's cache behaviour is unchanged** — the issue explicitly defers a per-key lock, so this is measurement only.

## The harness

The disk cache is off by default and no environment variable enables it, so each worker constructs its own `FileSystemJitStore` over its own directory per arm. The coordinator starts **only already-built binaries**, so no build or dependency resolution can land inside a timed region. A ready/start barrier makes every rank begin its first JIT+launch together, and `process_to_result` and `barrier_to_result` are recorded separately.

Two choices are worth calling out because the obvious alternative is wrong for both:

* **`tileiras` spawn attempts come from a pass-through wrapper, not `jit_backend_compile_count()`.** That counter is incremented only after a non-empty cubin is produced, so it is a *success* count: it cannot distinguish "the compiler never ran" from "the compiler ran and failed", and reporting it as a spawn count would be wrong.
* **The L2 key is derived up front through the frontend alone, from meta tensors**, so the pre-check runs no backend and no GPU. Rounds whose ranks disagree on the key are recorded but excluded from same-key conclusions.

`RUSAGE_CHILDREN` is added to `RUSAGE_SELF`, so child CPU time covers the `tileiras` subprocess rather than just the Rust parent. Every arm uses the same wrapper, and `--version` is passed through untouched so the compiler fingerprint — part of the key — is unchanged.

## What it measures

`barrier_to_result_ms` is the contention-sensitive number: everything before the barrier is identical in every arm. `jit_attributable_ms` subtracts the `relaunch_min` floor (release → all ranks correct, ms; median, 10 trials/config, disk cache disabled / per-process cold private dir / shared cold dir / single-process warm then shared warm dir):

| kernel | N | cache disabled | private cold | shared cold | warm shared |
|---|---|---|---|---|---|
| add | 1 | 1285.2 | 1298.8 | 990.5 | **425.1** |
| add | 4 | 1315.6 | 1882.7 | 3134.7 | **317.0** |
| add | 8 | 1190.4 | 1220.0 | 4634.6 | **458.7** |
| gemm | 4 | 1521.9 | 1333.3 | 2817.7 | **404.7** |
| gemm | 8 | 1456.6 | 2203.8 | 4571.7 | **471.3** |

Duplicate `tileiras` spawns per round (sum over ranks) are exactly N for the three cold arms and 0 for the warm arm, for every N and both kernels.

## Findings a reviewer should read before building on this

**The stampede is real and deterministic.** A shared cold directory at N ranks produced exactly N spawns of **one** key, N successful compiles, N puts, 0 hits, 0 I/O errors, and exactly **one** `<key>.cubin` on disk whose name equals the independently derived meta key. 10/10 trials at N=1,2,4,8 for both kernels. The identical counts also hold with the cache disabled and with per-process private directories, so **the redundancy is per cold process, not an artefact of sharing a directory** — there is no lock convoy to fix, only duplicated work.

**The penalty is not only redundant compilation.** A clean phase split (add, N=8, n=80/arm, 0 dropped) shows per-rank CPU is *identical* across arms (378 / 382 / 382 ms), while post-compile time — cubin read, encode, store put, module load, launch — is 19 / 22 / **247** ms for cache-disabled / private / shared. The extra time is **blocking, not computing**, and `strace` of the same configuration puts it in the single shared shard directory: the worst single `rename` into it took **1294 ms** against 0.1 ms in the private arm (worst `openat` 848 ms against 397 ms). That filesystem share of the penalty belongs to this host and a compile lock would not remove it. Both are reproducible with `scripts/jit_stampede_probe.py <raw-dir> --kernel add --p 8 --tag phases --phases`.

**Consequence:** a future compile lock would remove the redundant compiles *and* replace the losers' `put` with a cheaper `get`, but the filesystem component of the penalty is environment-specific and would not follow the fix.

**The documented warmup workaround already satisfies the need.** At N=8 its total cost including the warmup itself (add 2330 ms, gemm 3426 ms) is *below* shared-cold startup alone (4635 / 4572 ms), and post-warm startup (290–471 ms) is below even the cache-disabled baseline (1190 / 1457 ms).

## Not measured — read these as scope limits

* **One-process-per-GPU was covered by a supplement, not by the primary run.** GPUs 4 and 5 are held in `Exclusive_Process` compute mode by another user's vLLM workers (42.9 GB each; context creation returns `CUDA_ERROR_DEVICE_UNAVAILABLE 46`), so every rank of the primary run shared GPU 3, each rank independently confirming `/dev/nvidia3` via `/proc/self/fd` and recording the expected UUID. A re-run with the ranks round-robined over two distinct devices is in `artifacts/205/raw/multigpu-summary.md`: the duplicate compilation is unchanged, but at `add`/P=4 the warm-cache advantage falls from 9.9× to 2.5×, because much of the single-device penalty is contention rather than redundant compilation. **The single-device magnitudes above must be read together with that correction.**
* **All wall-clock numbers are conditioned on a heavily loaded shared host** (load average 130–1370). The shared-cold magnitude drifted between 1519 ms and 4635 ms across runs. The multi-second figures must not be read as hardware-general.
* Also unmeasured: an idle-machine or low-latency-filesystem baseline, NFS/multi-host sharing, any lock-based arm, how much of the directory-metadata cost a lock would remove, larger kernels, and 30 trials for every configuration (a replication run at higher load reached 13–14 trials before being stopped).

## Rebase re-measurement

The branch was rebased onto `main` (`e04245b`), so the matrix was re-run there rather than carrying the primary run's numbers across a changed base. 144 rounds (24 configurations × 5 trials, plus one correctness round each), **0 invalid**, every rank's derived key matching the key the runtime wrote:

| kernel | P | shared cold | warm shared | warm advantage | `tileiras` attempts per round (shared) |
|---|---|---|---|---|---|
| add | 1 | 789.8 ms | 261.0 ms | 3.0× | 1 |
| add | 2 | 1136.2 ms | 266.4 ms | 4.3× | 2 |
| add | 4 | 1181.3 ms | 283.2 ms | 4.2× | 4 |
| add | 8 | 2143.3 ms | 352.0 ms | 6.1× | 8 |
| gemm | 1 | 1742.4 ms | 350.7 ms | 5.0× | 1 |
| gemm | 2 | 1909.6 ms | 356.3 ms | 5.4× | 2 |
| gemm | 4 | 2380.6 ms | 397.9 ms | 6.0× | 4 |
| gemm | 8 | 3661.0 ms | 421.3 ms | 8.7× | 8 |

Every conclusion reproduces on the new base: the spawn histogram is exactly `{"P": 5}` in all 5 trials of every shared configuration and `{"0": 5}` in every warm one, so the redundancy is one compile per cold process; the warm arm stays flat in P (261–421 ms) while the cold arm grows; and the warm flow's total remains below the cold start. The host was more loaded for this run (load average median 845 against 551 for the primary run), so the absolute milliseconds are lower than the primary tables — the comparable quantity is the ratio, and the direction is the same.

## Re-running

```bash
source env-cutile.sh   # CUDA/tileiras/LD_LIBRARY_PATH for the host
scripts/run_jit_stampede_205.sh --trials 10 --correctness-rounds 3 \
  --p 1,2,4,8 --modes nocache,private,shared,shared-warm \
  --kernels add,gemm --devices 3 --tag run1
```

`scripts/jit_stampede_tables.py` re-derives the README tables from `summary.json`; `scripts/jit_stampede_probe.py` produces the per-phase splits. `artifacts/205/README.md` carries the environment fingerprint, the rendered tables, the phase split and the limitations; its appendix re-measures the coordinator's strict-parsing edit with the two versions alternated inside each repetition, and shows the edit moves no measured number.

Refs #205 — Phase 1 only. The issue's second phase (bounded per-key coordination with a redundant-compile fallback) is **not** implemented here, and no claim is made that the stampede is fixed.


0z5a:
Closing the main gap flagged in the description: the primary run placed every rank on one device, so **one-process-per-GPU was never exercised**.

I re-ran with ranks assigned round-robin across two distinct physical devices (`--devices 1,3`, 13 rounds/config, 0 invalid). Each rank independently confirmed its device and recorded the UUID:

```
rank 0 -> physical GPU 1    uuid GPU-739e6fcb-…
rank 1 -> physical GPU 3    uuid GPU-dfe309b3-…
rank 2 -> physical GPU 1
rank 3 -> physical GPU 3
```

**The duplicate-compile conclusion holds.** Sum of spawn attempts over all ranks, per round: exactly P for `nocache` and `shared`, and exactly 0 for `shared-warm` — so sharing one cold cache directory *across two devices* still produces one compile per cold process. The redundancy is per cold process, not an artefact of a single device. (One of 13 `shared`/`add`/P=4 rounds saw an early-finisher dedup: 3 attempts instead of 4.)

**The magnitude does not hold,** and the direction is worth stating plainly because it argues *against* the headline number:

| kernel | P | single-GPU shared | multi-GPU shared | single-GPU warm | multi-GPU warm |
|---|---|---|---|---|---|
| add | 4 | 3134.7 ms | **762.1 ms** | 317.0 ms | **305.5 ms** |
| gemm | 4 | 2817.7 ms | **1340.5 ms** | 404.7 ms | **393.1 ms** |

The warm-cache advantage is 2–3× here against roughly 10× in the primary run. That is exactly what the phase analysis predicts: much of the single-device penalty was contention on one saturated device and its disk, not the redundant compilation. With the load spread across devices, the saving is mostly compiler work and CPU time, and the absolute wall-clock latency is already only 0.7–1.4 s.

**Practical consequence: the general-case wall-clock benefit is smaller than the primary run suggests.** Anyone weighing a fix should treat the ~10× figure as an artefact of that specific contended host, and the honest summary as: the redundant work is real and scales with process count, the wall-clock win depends heavily on the filesystem and device contention, and the documented warmup workaround already covers the launcher that can afford one warmup process.

Evidence committed in `artifacts/205/raw/multigpu-summary.md`.

PR #300, opened 2026-09-21
Add bytecode reader. Allows cutile-rs to run existing cuda-tile kernels

ivarflakstad:
With `cutile_ir::read_bytecode` we can now deserialize Tile IR bytecode into a `cutile_ir::Module`. Which means users that already have existing cuda-tile kernels can easily use these from cutile-rs (https://docs.nvidia.com/cuda/cutile-python/compilation.html).

I am no expert in the details of the Tile IR bytecode format, so instead I've relied on correctness from round-trip byte-equality (`write(m) -> read -> m' -> write(m'), m == m'`). There may be cases I've missed, but I don't think it'll be easy to find those from code review. Imo it would be much more efficient to run the round-trip on a bunch of existing bytecode, which I'm sure Nvidia has laying around :)
I added a round-trip smoke test to `cutile-kernels` as well.

When I first added the reader I also added a bunch of duplication. It might have made for a simpler first review, but in the long run I think re-using the concepts between reader/writer/etc is better. That's mostly why there are a bunch of changes.

PR also fixes the todo wrt decoding optimization hints in `parse_func_section`.

Tried to keep to the style of the repo.

Feel free to make any changes 👍


elibol:
@ivarflakstad, love this.

I caught one issue: Reading real cutile-python bytecode (test/bytecode_dump fixtures) fails on every pre-13.3 file: the reader expects the 13.3+ leading optional-flags bitfield in partition_view, but pre-13.3 the layout is tile_shape / tensor_view / dim_map with a trailing has-padding varint (see kUnifiedBitfieldVersion in cuda-tile and cutile-python's TypeTable.partition_view), and the round-trip tests can't catch it because the writer mirrors the same assumption. Worth adding those fixtures as a cross-implementation read test.

PR #306, opened 2026-09-22
perf: cut the per-launch host cost of access tracking on the await path

elibol:
Follow-up to #302 for grout's remaining short-prefill regression (+4.3% on pp18, host-bound; ~0.5 µs per awaited op on 6–10-argument kernels). Cross-stream conflict detection and lifetime retention are unchanged.

- A scheduled `DeviceFuture` reuses the context's own submission when nothing else shares it, instead of allocating a second one per awaited op.
- In-flight tensor accesses live in eight lock-free packed slots per storage (`used|write|recording|stream id|serial`) with a mutex overflow list; the access is claimed before the conflict scan, so of two concurrent conflicting acquirers at least one observes the other. `Stream::id()` (interned per handle) makes the stream fit the word exactly.
- The generated launcher collects its arguments' leases into one batch and retains them under a single lock (`KernelInputStored`/`KernelOutputStored::retain` take the batch).

Host-only enqueue cost (RTX 5090, `benches/host_overhead`, 4-tensor kernel): async_on launch 1.897 → 1.825 µs, awaited launch 6.959 → 6.877 µs (pre-#275: 1.617 / 6.663). What remains is one submission allocation, one `Arc` inc/dec and two CASes per argument, and one lock per launch.


elibol:
Closing: the per-launch access records this makes lock-free are being removed outright. A launch now holds one Arc per tensor allocation and nothing else, so there is nothing left here to optimize. grout measured this branch as neutral.

